PR #8, opened 2025-02-24
SNT-17 [Front] As a user I can manage scenarios

bramj:
Not perfect, but good enough for a first version. The buttons on the top right (export/duplicate) are fake for now.

[scenarios-2025-02-24_12.25.07.webm](https://github.com/user-attachments/assets/7a90fde7-c3d0-4005-a289-c8224d8e1ed8)


hakifran (commented):
image
There is a warning in the console.

hakifran (approved):
It works fine but there are some improvement to make on the code. But this can be merged as we are in rush

PR #39, opened 2025-06-27
Snt 57 update colors for covariants and layouts on map

Bewi:
Updating colors based on this : 
https://bluesquare.atlassian.net/browse/SNT-57

For ranges, calculate 7 steps for the legend to have something a bit smoother than only two.

This is how it looks on RDC:
https://files.slack.com/files-pri/T02T1R13H-F0938PL8KFW/image.png

### DIFF
--- a/management/commands/support/legend.py
+++ b/management/commands/support/legend.py
@@ -53,8 +53,7 @@ def get_legend_config(metric_type):
     max_value = result["max_value"]
 
     if metric_type.category == "Mortality":
-        steps = get_steps(0, max_value, len(SEVEN_SHADES))
-        return {"domain": steps, "range": SEVEN_SHADES}
+        return {"domain": [0, max_value], "range": [NINE_SHADES[0], NINE_SHADES[-1]]}
     if metric_type.category == "Composite risk":
         return {
             "domain": list(range(int(min_value), int(max_value))),
--- a/management/commands/support/legend.py
+++ b/management/commands/support/legend.py
@@ -84,9 +84,6 @@ def get_steps(min, max, count):
     round_digits = 2 if min < 1 else 0
     if count == 1:
         return [round(min, round_digits)]
-    print(f"{min}, {max}, {count}")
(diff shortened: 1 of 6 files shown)


management/commands/support/legend.py
```diff
@@ -53,7 +53,8 @@ def get_legend_config(metric_type):
     max_value = result["max_value"]
 
     if metric_type.category == "Mortality":
-        return {"domain": [0, max_value], "range": ["#FFCCBC", "#601B06"]}
+        steps = get_steps(0, max_value, len(SEVEN_SHADES))
+        return {"domain": steps, "range": SEVEN_SHADES}
```
bramj, line 57: I think we might want to keep only 2 colors here, the mortality is a bit special in that it's a linear scale, see lines `74-75` of this same file.
The resulting legend is a gradient color going from color 1 to color 2. So I propose we use the first and last of the new colors like 
```python
return {"domain": [0, max_value], "range": [NINE_SHADES[0], NINE_SHADES[-1]]}
```

management/commands/support/legend.py
```diff
@@ -74,3 +76,14 @@ def get_legend_type(metric_type):
     if metric_type.category in ["Composite risk", "Seasonality"]:
         return "ordinal"
     return "threshold"
+
+
+def get_steps(min, max, count):
+    if not min or not max:
+        return []
+    round_digits = 2 if min < 1 else 0
```
bramj, line 84: :+1:

management/commands/support/legend.py
bramj: (at least for Burkina Faso that's the case, for RDC we don't have the data yet. But I think it's good to keep an example of a linear legend in the script)

bramj (commented):
Nice work! :clap: This is looking exactly like I had in mind. I think there are some things to improve when values are missing for example, but let's see what feedback we get from this.
Just a minor comment.

PR #40, opened 2025-06-30
Snt 58 district detail popup

Bewi:
Story: https://bluesquare.atlassian.net/browse/SNT-58

- Layout for org unit details on map.
- Selection from filter is not linked to selection on the mix.
- When one item is selected from map, property display add to / remove from mix.
- When removing from mix directly, update map state.
- Selected items on map are now related to filter and not selection on mix.

https://github.com/user-attachments/assets/3ab55bb8-84b0-46cf-a4a6-edbd4728416d

### DIFF
--- a/js/src/domains/planning/components/MapOrgUnitDetails.tsx
+++ b/js/src/domains/planning/components/MapOrgUnitDetails.tsx
@@ -74,7 +74,6 @@ const styles: SxStyles = {
     }),
     closeIconButton: (theme: Theme) => ({
         color: theme.palette.text.primary,
-        // paddingLeft: 0,
         paddingRight: 0,
     }),
     closeIcon: {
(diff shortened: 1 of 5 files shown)


js/src/domains/planning/components/MapOrgUnitDetails.tsx
```diff
@@ -22,58 +22,64 @@ import {
...
+    title: (theme: Theme) => ({
+        marginRight: theme.spacing(1),
+        flexGrow: 1,
+        whiteSpace: 'nowrap',
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
+    }),
+    closeIconButton: (theme: Theme) => ({
+        color: theme.palette.text.primary,
+        // paddingLeft: 0,
```
bramj, line 77: forgotten comment

bramj (approved):
That's looking really nice and working well :clap: 

As a minor comment (take it or leave it), I think a `min-width` to fit the "districts selected" button would be nice:

For example here it would give a bit more breathing room to the org unit name, and would look nicer.

js/src/domains/planning/components/MapOrgUnitDetails.tsx
Bewi: I agree for the min width comment.
Thing is that the filter width changes based on message in it.
If you look at it when nothing is filtered, width are aligned: 
image

PR #45, opened 2025-07-09
Snt 76 soft delete scenarios

Bewi:
Soft delete scenarios from scenario list and planning page.
[SNT-76](https://bluesquare.atlassian.net/browse/SNT-76)

https://github.com/user-attachments/assets/892b376c-e23b-42d7-a60f-a5b6c9f00ee6

### DIFF
--- a/js/src/domains/scenarios/hooks/useGetScenarios.tsx
+++ b/js/src/domains/scenarios/hooks/useGetScenarios.tsx
@@ -82,10 +82,13 @@ export const useDuplicateScenario = (
         snackErrorMsg: MESSAGES.duplicateError,
     });
 
-export const useDeleteScenario = () =>
+export const useDeleteScenario = (
+    onSuccess: (data: any) => void = _data => null,
+): UseMutationResult =>
     useSnackMutation({
         mutationFn: (id: number) =>
             deleteRequest(`/api/snt_malaria/scenarios/${id}`),
+        options: { onSuccess },
         invalidateQueryKey: ['scenarios'],
         snackSuccessMessage: MESSAGES.deleteSuccess,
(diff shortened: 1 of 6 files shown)


bramj (changes requested):
Looking good, but we definitely need a confirmation modal when we hit delete on a scenario.

Also, can you check with Yannick/Jerome on the UI? I think for example maybe we only want the delete button on the list page, but not on the details.

Bewi:
After a chat with Jerome, it occurred that it does make sense to have the delete button on details, so you know what you are deleting.
On the list, it might not be obvious and they might remove the wrong element.

js/src/domains/scenarios/hooks/useGetScenarios.tsx
```diff
@@ -68,3 +81,15 @@ export const useDuplicateScenario = (
         snackSuccessMessage: MESSAGES.duplicateSuccess,
         snackErrorMsg: MESSAGES.duplicateError,
     });
+
+export const useDeleteScenario = (
+    onSuccess: (data: any) => void = _data => null,
+): UseMutationResult =>
+    useSnackMutation({
+        mutationFn: (id: number) =>
+            deleteRequest(`/api/snt_malaria/scenarios/${id}`),
```
bramj:
```suggestion
            deleteRequest(`/api/snt_malaria/scenarios/${id}/`),
```

bramj (approved):
LGTM :+1: 

I added 1 small commit with a few fixes (also unrelated to your changes, I couldn't add suggestions on this).

Not sure why I keep having the icon in grey (I see it looks correct in your video), so maybe it's just me
image

PR #46, opened 2025-07-11
Add page allow manual run of the import_openhexa_metrics Django command

bramj:
This adds a "hidden" page to `/snt_malaria/import_openhexa_metrics/` that allows an admin user to manually launch the Django command to import the metrics into a specific account.

Thanks to Claude Code for the nice interface :pray: 

image

image


Bewi (approved):
All is working fine and it is properly protected behind an account login !

Note: We should add this to the readme

PR #53, opened 2025-07-16
SNT-87: Fix ordinal scale color mapping 

Bewi:

Related JIRA tickets : [SNT-87](https://bluesquare.atlassian.net/browse/SNT-87)

- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] Are there enough tests?

## Doc

## Changes

When mapping ordinal metrics to color, we have mismatch if value is an actual number.
For this specific case, legend is using plain numbers (2, 3, 4) but values are formatted using floats (2.0, 3.0, 4.0).
Currently the comparison was done expecting two strings.
Changed that to compare as numbers if value is an actual number.

## How to test

If you have data for Congo, you can just go to a scenario and check layer from group "ENVIRONMENTAL INDICATOR", Metric tope: "Length of Tansmission Season by Rainfall (ERA5)"

## Print screen / video

image

[SNT-87]: https://bluesquare.atlassian.net/browse/SNT-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/src/domains/planning/libs/map-utils.tsx
+++ b/js/src/domains/planning/libs/map-utils.tsx
@@ -27,14 +27,30 @@ export const shouldReverse = (threshold: ScaleDomainRange) => {
     return threshold.domain[0] > threshold.domain[threshold.domain.length - 1];
 };
 
+const getColorForOrdinalScale = (
+    value: string | number,
+    legend_config: ScaleDomainRange,
+) => {
+    const numericValue = Number(value);
+    let index: number;
+    if (Number.isNaN(numericValue)) {
+        index = legend_config.domain.indexOf(value as never);
+    } else {
+        index = legend_config.domain.findIndex(
+            (d: number | string) => Number(d) === numericValue,
+        );
+    }
+
+    return legend_config.range[index];
+};
+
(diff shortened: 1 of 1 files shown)


bramj (commented):
I'm just wondering if instead of adding this complexity to the frontend, wouldn't it be easier and safer to be smart at the moment of the import?

For example one of the problematic data layers is "Length of Tansmission Season by Rainfall (ERA5)" for Burkina Faso.
The legend config is:
```json
{"range": ["#A5D6A7", "#FFECB3", "#FFAB91"], "domain": ["3", " 4", " 5"]}
```
So there are spaces there + they are strings, it should be `[3,4,5]`.
Maybe being smarter at the source and correct it on the import?

Bewi:
Replaced by: SNT-87-fix-ordinal-color-mapping-v2

Bewi:
Replaced by: https://github.com/BLSQ/snt-malaria/pull/59

PR #60, opened 2025-07-22
SNT-90: Remove intervention list and link OU to Intervention directly

Bewi:

Related JIRA tickets : SNT-90

- [x] My migrations file are included
- [ ] Are there enough tests?

## Doc

You can find the new layout in Figma: 
https://www.figma.com/design/BdXVNGLm1nuLC9eNFdfA7n/SNT-Malaria?node-id=4290-17377&m=dev

## Changes

We used to group intervention and org unit into mixes. 
A mix could have multiple interventions and org units.

It appears, this was overkilled and what is needed in the end is to have a direct link from intervention to org units.
So one intervention can be linked to multiple org units.

## How to test

Go to a scenario, select org units on the map, add to the list.
Select some interventions and add click the add button.
It should appear on the right.
You can play around and do multiple assignments.

Also, to double check all translations. 
"mix" should not appear anymore.

## Print screen / video

image

## Notes

This doesn't cover some part of the layout.
Such as: 
- Intervention assignment map.
- Run budget button
- Layer select on intervention plan
- Showing all org units linked to an intervention
- Remove a org unit from an intervention

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/js/src/domains/planning/index.tsx
+++ b/js/src/domains/planning/index.tsx
@@ -71,7 +71,7 @@ export const Planning: FC = () => {
         );
     };
 
-    const handleAddMapOrgUnitsToList = () => {
+    const handleAddMapOrgUnitsToList = useCallback(() => {
         const newOrgUnits = selectionOnMap.filter(
             orgUnit =>
                 !selectionOnInterventionList.some(
@@ -89,7 +89,7 @@ export const Planning: FC = () => {
                 }),
             ),
         );
-    };
+    }, [selectionOnMap, selectionOnInterventionList, formatMessage]);
 
     const { data: displayedMetricValues, isLoading } = useGetMetricValues({
         metricTypeId: displayedMetric?.id || null,


beygorghor:
@tdethier a review on the python part is welcome here please

api/interventionassignments/views.py
```diff
@@ -42,37 +40,30 @@ def create(self, request, *args, **kwargs):
...
-        # create the created or selected intervention mix and link it the interventions
-        if selected_mix:
-            intervention_mix = selected_mix
-        else:
-            intervention_mix, created = InterventionMix.objects.get_or_create(
-                name=mix_name, account=self.request.user.iaso_profile.account, scenario=scenario
-            )
-            intervention_mix.interventions.add(*interventions)
-
+        print(
```
tdethier, line 48: I would probably use a logger instead of `print`, but I'm not sure of how things are done in SNT

migrations/0012_remove_interventionassignment_deleted_at.py
```diff
@@ -0,0 +1,16 @@
...
+    dependencies = [
+        ("snt_malaria", "0011_alter_interventionassignment_unique_together_and_more"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="interventionassignment",
+            name="deleted_at",
+        ),
+    ]
```
tdethier, line 16: Nitpick: it's best to avoid having multiple migration files, otherwise you end up with 300+ like in classic iaso :sweat_smile: 

I would probably migrate back both of them, delete both of them and regenerate a single file

models/intervention.py
```diff
@@ -50,28 +49,14 @@ def __str__(self):
...
+class InterventionAssignment(models.Model):
     class Meta:
         app_label = "snt_malaria"
-        unique_together = [["scenario", "org_unit", "intervention_mix"]]
+        unique_together = [["scenario", "org_unit", "intervention"]]
 
     scenario = models.ForeignKey(Scenario, on_delete=models.CASCADE, related_name="intervention_assignments")
     org_unit = models.ForeignKey(OrgUnit, on_delete=models.PROTECT)
-    intervention_mix = models.ForeignKey("InterventionMix", null=True, on_delete=models.PROTECT)
+    intervention = models.ForeignKey(Intervention, on_delete=models.PROTECT, blank=True, null=True)
```
tdethier, line 59: I don't know anything about the SNT business logic, but since `InterventionMix.interventions` was a m2m relationship, are you sure about using a `models.ForeignKey` here?

tdethier (commented):
I didn't test the code, since I don't know anything about SNT and don't have a local setup. 

I just read the python code, it looks good :+1: 
I left a few comments, but I let you decide what to do, you know better than I do

js/src/domains/planning/hooks/UseRemoveOrgUnitFromInterventionPlan.tsx
```diff
@@ -2,11 +2,28 @@ import { UseMutationResult } from 'react-query';
...
+            if (!interventionAssignmentIds) {
+                return Promise.resolve([]);
+            }
+
+            const promises = interventionAssignmentIds.map(
+                interventionAssignmentId =>
+                    deleteRequest(`${baseUrl}${interventionAssignmentId}/`),
+            );
+
+            return Promise.all(promises);
```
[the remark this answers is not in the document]
Bewi: That would be a good addition indeed.
There is one slight detail though, for get, message is applied to every request.
Here I wanted to have one for all queries.
It could be a way to configure the useSnackMutations. 
I'll add a story to experiment that

PR #62, opened 2025-07-29
Snt 81 provide a way to show all interventions on a combined map

Bewi:

Related JIRA tickets : SNT-XXX

- [ ] Are there enough tests?

## Changes

Update intervention plan map so: 
How it works at the moment:

  **1. All interventions is selected**

- Highlight all org units linked to an intervention for this scenario

- A new notion of “group” exists which is made of interventions linked to the same scenario.
So if a scenario A has intervention A & intervention B, a new group exists.
If scenario B has only intervention B, they will be highlighted differently
- Colors are “randomly” generated based on the number of groups of scenario but will always be the same if nothing changes (upon refresh and such), we don’t want new random color everytime we visit a scenario that didn’t change
- A tooltip is displayed on hover of org units, showing their names and interventions if any.

**2. A specific intervention is selected**
- No legend is shown.
- Color of highlighted org units is always the same
- Only the name of the org unit is displayed when hovering an org unit (district) on the map

See screenshot bellow

## How to test

On a Scenario, create some intervention plans. 
To do that: Select district on the main map, add them to the list.

Scroll down (this one is not obvious), you will see what you selected and interventions.
You can select some interventions and click add to plan button just above.

On intervention plans, you will see how many districts (org units) are attached to any intervention.
If you open the map (small map icon on top), you should see the map with a dropdown to choose for which intervention you want to see org units highlighted.
By default, it will be All intervention EXCEPT if there is only one choice possible, in which case, it'll show that intervention selected.

On the right, you'll have 

## Print screen / video

**All interventions is selected:** 

image

**A specific intervention is selected**

image

```
fix: empty instance pop up

Refs: IA-3665
```


js/src/domains/planning/messages.ts
```diff
@@ -128,4 +128,8 @@ export const MESSAGES = defineMessages({
         id: 'iaso.snt_malaria.label.searchPlaceholder',
         defaultMessage: 'Search District',
     },
+    allInterventions: {
+        id: 'iaso.snt_malaria.label.allInterventions',
```
bramj, line 132: Translations missing for this one I think?

js/src/domains/planning/components/interventionPlan/InterventionsPlanMap.tsx
```diff
@@ -123,50 +253,42 @@ export const InterventionsPlanMap: FunctionComponent<Props> = ({
...
-                </Select>
-            </Box>
-            <Box sx={styles.legendBox}>
-                <Box sx={styles.legendShape} />
-                <Typography variant="caption" sx={{ color: 'white' }}>
-                    {`${highlightedOrgUnits.length} ${formatMessage(MESSAGES.orgUnitDistrict)}`}
-                </Typography>
+                />
             </Box>
+            {selectedPlanId ? null : <MapLegend legendConfig={legendConfig} />}
```
bramj, line 291: There's a small issue with the bottom padding/margin here. Also, I think for this map in particular, it would be clearer to display them vertically stacked like a column, what do you think?

Image

js/src/domains/planning/components/interventionPlan/InterventionsPlanMap.tsx
```diff
@@ -123,50 +253,42 @@ export const InterventionsPlanMap: FunctionComponent<Props> = ({
...
+                                fillOpacity: 2,
+                            }}
+                        >
+                            <Tooltip>
+                                <b>{orgUnit.short_name}</b>
+                                {orgUnitMapMisc.label && (
+                                    <>
+                                        <br />
+                                        {orgUnitMapMisc.label}
+                                    </>
```
bramj, line 276: That's really nice 👍

js/src/domains/planning/libs/map-utils.tsx
```diff
@@ -1,9 +1,32 @@
...
+export const maxHue = 350;
+
+export const getColorRange = (count: number = 1) => {
+    const colorStep = Math.round(maxHue / count);
+    let prevHue = 0;
+
+    return Array.from(Array(count)).map(() => {
+        const [hue, sat, light] = getHslColor(prevHue, colorStep);
+        prevHue = hue;
+        return hslToRgb(`hsl(${hue},${sat * 100}%,${light * 100}%)`);
```
bramj, line 17: Very cool, nice colors 👏

bramj (changes requested):
Really nice @Bewi ! 👏 
The colors play out very nice, and along with the tooltip it gives a really nice map.

I wonder if we can make the "Intervention plan" section even a bit more wider, like 60%, but it's already great as is.

Just a minor remark on the legend styling.

quang-le (changes requested):
Very nice work. Works like a charm for me.
A few nitpicks code-style wise. 
I'm going to be annoying and block on the `return` inside the `forEach`loop and the `reduce` that should be a `forEach`.

About the feature, it would be nice to be able to switch the map to full screen. We have that in Iaso: https://github.com/BLSQ/iaso/blob/main/hat/assets/js/apps/Iaso/domains/registry/components/map/MapToggleFullscreen.tsx

js/src/domains/planning/components/interventionPlan/InterventionsPlanMap.tsx
```diff
@@ -83,24 +110,127 @@ export const InterventionsPlanMap: FunctionComponent<Props> = ({
...
+                .map(i => i.id)
+                .join('--');
+
+            // We don't want to generate new color if one was already assigned for this group.
+            const existingMap = colorMap.find(
+                c => c.interventionsKey === interventionGroupKey,
+            );
+            if (existingMap) {
+                existingMap.orgUnitIds.push(key);
+                return;
```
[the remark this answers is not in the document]
Bewi: I will use a .reduce instead, which fits better with what this is actually doing

js/src/domains/planning/components/interventionPlan/InterventionsPlanMap.tsx
Bewi: That's something I discussed with Jerome already, plan is to keep it like this for now and see with the usage what we want to do. 
Vertically, the problem will be the overflow (scrollable legend), even though this means a lot of different groups of intervention

quang-le (approved):

bramj (approved):

PR #71, opened 2025-08-21
SNT-109 Set up account for RDC and BFA

Bewi:

Related JIRA tickets : SNT-109

- [ ] Did I use eslint and ruff formatters?
- [ ] Is my code clear enough and well documented?
- [ ] Are my typescript files well typed?
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] Are there enough tests?

## Doc

## Changes

We already have a script to create a dummy account with dummy data.
This is to create a multi tenant user with an account based on fixed keys.

It will create a user account if it doesn't exist with "admin" password  (should be changed by the user asap)
Example of usage for BFA: `manage.py set_up_snt_account --account_config_name BFA --main_username UserName`
Example of usage for RDC: `manage.py set_up_snt_account --account_config_name RDC --main_username UserName`

Once done, it will suggest to run another command to import metric values, this should be run as well.

## How to test

From a new DB, once the script has run.
You should be able to login, see the account you run the script for.
Try to create a scenario, you should see the map with metrics on it.

To go further, try to run the script for both BFA and RDC for the same user.
It should work and you should be able to change account freely.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


management/commands/setuper.py
```diff
@@ -0,0 +1,159 @@
...
+                first_name=main_user.first_name,
+                last_name=main_user.last_name,
+                email=main_user.email,
+                is_superuser=main_user.is_superuser,
+                is_staff=main_user.is_staff,
+            )
+
+            # Link main user to account
+            tenant_user = TenantUser.objects.create(main_user=main_user, account_user=account_user)
+            profile, _ = Profile.objects.get_or_create(user=tenant_user.account_user, account=account)
```
tdethier, line 96:
```suggestion
            profile, _ = Profile.objects.get_or_create(user=account_user, account=account)
```

management/commands/setuper.py
```diff
@@ -0,0 +1,159 @@
...
+            self.stdout.write(
+                self.style.SUCCESS(f"Linked main user '{main_user.username}' to account '{account.name}'")
+            )
+
+            # Create Project
+            project = Project.objects.create(name=project_name, account=account, app_id=uuid.uuid4())
+            profile.projects.add(project)
```
tdethier, line 103: Doing this means that this user will not have access to other projets that could be created in that account. I'm not sure if there will be multiple projects in a single account in SNT, but just FYI in case you missed it.

management/commands/setuper.py
```diff
@@ -0,0 +1,159 @@
...
+                    version_number=source_version.number,
+                    validation_status="VALID",
+                    user=main_user,
+                    description=f"{account.name} organizational units imported from GPKG",
+                    task=None,
+                )
+                self.stdout.write(
+                    self.style.SUCCESS(f"Successfully imported {total_imported} organizational units from GPKG")
+                )
+                source_version.orgunit_set.update(validation_status="VALID")
```
tdethier, line 145: I was going to say that you don't need to do that, since you're passing `validation_status="VALID"` to `import_gpkg_file2()`, but there might be a bug with this `validation_status`

management/commands/setuper.py
```diff
@@ -0,0 +1,159 @@
...
+
+            # Import org units from GPKG
+            self.stdout.write(f"Importing org units from GPKG file: {gpkg_path}")
+            try:
+                total_imported = import_gpkg_file2(
+                    filename=gpkg_path,
+                    source=data_source,
+                    version_number=source_version.number,
+                    validation_status="VALID",
+                    user=main_user,
```
tdethier, line 138:
```suggestion
                    user=account_user,
```

management/commands/setuper.py
```diff
@@ -0,0 +1,159 @@
...
+                defaults={
+                    "email": f"{main_username}@bluesquarehub.com",
+                    "first_name": "Admin",
+                    "last_name": "User",
+                    "is_staff": True,
+                    "is_superuser": True,
+                },
+            )
+            if created:
+                main_user.set_password("admin")
```
tdethier, line 71: I'm not a big fan of that, in case it ends up being used in prod and no one thinks of changing this password. Maybe we could generate a random one and print it when the command ends (with a warning message telling the user to login and change it, because it was randomly generated and will be lost)?

tdethier (changes requested):
The feature works fine :+1: 

There are a few nitpicks (I made some suggestions here and there), but my main concern is the main user password. We can discuss that if you want.

management/commands/setuper.py
Bewi: I think those accounts will only have one project, for the region (RDC, BFA, ...) 
But good point, to be honest, I am not sur yet on how else to do that, not adding it to the profile ?

tdethier: Yeah, not adding any project = having access to all of them.
In SNT, the result is probably the same :shrug:

Bewi: I'll try that

tdethier (approved):
LGTM :+1:

PR #89, opened 2025-09-16
SNT-139 Add script to facilitate importing health pyramids of new countries

bramj:
Setting up new countries is a bit annoying at the moment.

This PR adds a Django management command to transform health pyramid parquet file and GeoJSON to GeoPackage format.
The result is a GeoPackage (.gpkg) file compatible with IASO.

Example usage:

```bash
./manage.py transform_health_pyramid \
    --pyramid-file plugins/snt_malaria/NER_pyramid.parquet \
    --geojson-file plugins/snt_malaria/NER_shapes.geojson \
    --output-file plugins/snt_malaria/NER_health_pyramid.gpkg
```

**This script served well for both Niger as well as Nigeria. For now we won't add it to the readme, I just want it to live in the repo so we keep track of it.**

### DIFF
--- a/management/commands/transform_health_pyramid.py
+++ b/management/commands/transform_health_pyramid.py
@@ -89,8 +89,35 @@ class Command(BaseCommand):
                 filtered_pyramid,
                 left_on="ADM2_ID",
                 right_on="LEVEL_3_ID",
-                how="left",
+                how="inner",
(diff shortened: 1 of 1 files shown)


Bewi (approved):
LGTM

PR #92, opened 2025-09-17
Snt 127 edit detailed cost per intervention

Bewi:
Handle intervention cost details

Related JIRA tickets : SNT-127

- [x] My migrations file are included
- [ ] Are there enough tests?

## Doc

## Changes

Explain the changes that were made.

The idea is not to list exhaustively all the changes made (GitHub already provides a full diff), but to help the reviewers better understand:

- which specific file changes go together, e.g: when creating a table in the front-end, there usually is a config file that goes with it
- the reasoning behind some changes, e.g: deleted files because they are now redundant
- the behaviour to expect, e.g: tooltip has purple background color because the client likes it so, changed a key in the API response to be consistent with other endpoints

## How to test

Explain how to test your PR.

If a specific config is required explain it here: dataset, account, profile, etc.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


admin.py
```diff
@@ -61,3 +63,19 @@ class ScenarioAdmin(admin.ModelAdmin):
...
+@admin.register(InterventionCostCategory)
+class InterventionCostCategoryAdmin(admin.ModelAdmin):
+    list_display = ("name", "id", "created_by", "created_at")
+    search_fields = ("name", "id")
+    list_filter = ("name", "id")
+    ordering = ("id",)
+
+
+@admin.register(InterventionCost)
+class InterventionCostCategoryAdmin(admin.ModelAdmin):
```
tdethier, line 77:
```suggestion
class InterventionCostAdmin(admin.ModelAdmin):
```

api/interventioncosts/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+
+class InterventionCostSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = InterventionCost
+        fields = ["id", "name", "cost", "category_id"]
+        read_only_fields = fields
+
+
+class InterventionCostWriteSerializer(serializers.ModelSerializer):
+    intervention_id = serializers.IntegerField(write_only=True)
```
tdethier, line 14: I would use a `serializers.PrimaryKeyRelatedField` instead, which will automatically handle parsing and checking object existence for you - you'll be able to remove lines 27-33

api/interventioncosts/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+    costs = serializers.ListField(child=serializers.DictField())
```
tdethier, line 15: What is the format of `costs`? Wouldn't it be better to define some kind of custom serializer for a single cost element, and then use your new serializer here for `costs` with `many=True`?

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_queryset(self):
+        return InterventionCost.objects.prefetch_related("intervention").filter(
+            intervention__intervention_category__account=self.request.user.iaso_profile.account
+        )
+
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        print(f"{serializer.validated_data}")
```
tdethier, line 35:
```suggestion
```

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Get validated objects
+        intervention = serializer.validated_data["intervention"]
+        costs = serializer.validated_data["costs"]
+        cost_categories = serializer.validated_data["cost_categories"]
+        print(cost_categories)
```
tdethier, line 41:
```suggestion
```

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+        cost_categories_dict = {value.id: value for _, value in enumerate(cost_categories)}
+        print(f"{costs}")
```
tdethier, line 43:
```suggestion
```

api/interventioncosts/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    class Meta:
+        model = InterventionCost
+        fields = ["intervention_id", "costs"]
+
+    def validate(self, attrs):
+        super().validate(attrs)
+        request = self.context.get("request")
+        account = request.user.iaso_profile.account
```
tdethier, line 24: Are you sure that all users will be logged in + will have a iaso profile? Otherwise this will probably crash

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class InterventionCostsViewSet(viewsets.ModelViewSet):
+    ordering_fields = ["id"]
+    http_method_names = ["get", "options", "post"]
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = InterventionCostListFilter
+
+    def get_serializer_class(self):
+        if self.request.method == "POST":
```
tdethier, line 22:
```suggestion
        if self.action == "create":
```

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+            return InterventionCostWriteSerializer
+        return InterventionCostSerializer
+
+    def get_queryset(self):
+        return InterventionCost.objects.prefetch_related("intervention").filter(
```
tdethier, line 27:
```suggestion
        return InterventionCost.objects.select_related("intervention").filter(
```

api/interventioncosts/views.py
```diff
@@ -0,0 +1,64 @@
...
+        serializer.is_valid(raise_exception=True)
+
+        print(f"{serializer.validated_data}")
+
+        # Get validated objects
+        intervention = serializer.validated_data["intervention"]
+        costs = serializer.validated_data["costs"]
+        cost_categories = serializer.validated_data["cost_categories"]
+        print(cost_categories)
+        cost_categories_dict = {value.id: value for _, value in enumerate(cost_categories)}
```
tdethier, line 42:
```suggestion
        cost_categories_dict = {value.id: value for value in cost_categories}
```

tdethier (changes requested):
Some comments:
- I only looked at the code, I did not test the feature yet
- If you can't create `InterventionCostCategory` objects through the API, shouldn't you have some kind of migration to provide the base items required for this feature?
- I added a few suggestions here and there, mostly improvements but a few nitpicks
- I did not understand the `InterventionCost` creation process through the API, but I don't know the business logic behind this feature - I think we should discuss that together
- I don't see any tests for these new models/API endpoints - you should really add some

models/intervention.py
```diff
@@ -42,6 +42,7 @@ class Meta:
         decimal_places=2,
         null=True,
         blank=True,
+        default=0
```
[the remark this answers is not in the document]
Bewi: You can see Intervention as a "global cost", which define the unit ("per 100 case" for example). 
And the "total cost" or "cost_per_unit". 

InterventionCost is a breakdown of the cost_per_unit and is not mandatory.
So you can have an Intervention without any InterventionCost and still having a cost_per_unit.

api/interventioncostcategories/serializers.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework import serializers
```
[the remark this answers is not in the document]
Bewi: Yeah, I followed all names format we have in api for snt_malaria.
If I renamed this, I need to rename all others

bramj (changes requested):
I think we're a bit on the wrong track here and we should discuss the general architecture.

I think the main issues are:
- the "cost category" can be simplified and (I believe) a hard-coded enum allowing these 4 options: Procurement, Distribution, Operational, Supportive
- the detailed cost is missing the `year` attribute (I guess this is missing from the mockup as well)

Looking at the PATH budgeting:
https://github.com/PATH-Global-Health/malaria-budgeting-tool/blob/main/www/user_inputs.md?plain=1#L43-L48
you can see that we currently don't have the following attrs: `local_currency_cost`, `usd_cost`, `cost_year`, `source`.
(I believe we will improve their script to avoid having to set the local currency, and we can take care of that later. Not sure if the `source` is used in the calculations, but we can also look at that later).

But my main issue is that right now we allow free-text addition of any type of cost category, which will not work with the budgeting script.
So I propose to fix the values on the ones from PATH.

models/intervention.py
```diff
@@ -60,4 +54,3 @@ class Meta:
     intervention = models.ForeignKey(Intervention, on_delete=models.PROTECT)
     created_by = models.ForeignKey(User, on_delete=models.PROTECT)
     created_at = models.DateTimeField(auto_now_add=True)
-    updated_at = models.DateTimeField(auto_now=True)
```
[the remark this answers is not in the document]
Bewi: Indeed, this was a mistake, I restored it and added updated_by

tdethier (commented):
Thanks for the changes, it's much better now 👍 

I'll let @bramj review and approve

bramj (approved):

PR #251, opened 2026-04-01
Fix SNT-410: org unit filtering for interventions

michiwend:

Scopes rule matching, CSV export/import, and form dropdowns to org units with configurable `intervention_org_unit_type`

### Related JIRA tickets

SNT-410

## Changes

The pre-existing field `intervention_org_unit_type` was used in a misleading way for filtering/zooming the map in the front-end. This PR repurposes this field for scoping org units to an org unit type that is relevant for intervention assignments so rules, imports/exports and form dropdowns contain only the relevant org units. The field is optional and the logic falls back to all org units when not provided.
For the purpose of filtering/zooming the map, a new field is introduced `focus_org_unit_type` which communicates the purpose more clearly. The migration moves existing configuration to the new field.

## How to test

* Create a match-all rule, go the the scenario comparison and look at the warnings, you should see warnings for unmatched org units on a higher level
* Configure the `intervention_org_unit_type` to the correct district level
* Repeat the first step and see that those org units are gone from the warning

Wrong org units included by a match-all appear in the impact warnings
image

Example configuration for Cameroon
image

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -195,6 +195,6 @@ class ImpactOrgUnitMappingAdmin(admin.ModelAdmin):
 
 @admin.register(AccountSettings)
 class AccountSettingsAdmin(admin.ModelAdmin):
-    list_display = ("id", "account", "intervention_org_unit_type_id")
+    list_display = ("id", "account", "focus_org_unit_type_id", "intervention_org_unit_type_id")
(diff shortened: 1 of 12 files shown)


Bewi (approved):
WaE & LGTM

PR #253, opened 2026-04-02
SNT-351 Cleanup planning V1

Bewi:

Cleaning up files from old version of planning

### Related JIRA tickets

SNT-351

## Changes

Explain the changes that were made.

Removed all unused files from planning V1 and renamed planningV2 to planning.

## How to test

Regression test, go to SNT and verify that nothing is broken with the Scenario, Compare & Optimize and Data Layers

## Print screen / video

N/A


michiwend (approved):
LGTM and tested locally

PR #256, opened 2026-04-09
[SNT-298] SNT account setup - backend

tdethier:

Backend for letting public users create accounts in SNT

### Related JIRA tickets

SNT-298

## Changes

- Added a new endpoint `/api/snt_malaria/account_setup/`
- Added tests for the new serializer & view
- Updated the intervention seeder in order to remove prints if not requested

## How to test

- There's no frontend yet, so the only way to test this is to:
  - manually make calls to the new endpoint (check unit tests to see examples)
  - or use the DRF web view
- You'll need to activate the feature through an env var `ENABLE_PUBLIC_ACCOUNT_SETUP`
- You can use the `tests/fixtures/geo_json_be.json` file for testing, it should be valid

## Print screen / video

image

image

## Notes

- you need [a IASO PR](https://github.com/BLSQ/iaso/pull/2900) in order to run this

## Doc

/


api/account_setup/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+            )
+
+        return value
+
+    def validate(self, data):
+        password = data["password"]
+        password_confirmation = data["password_confirmation"]
+
+        if password != password_confirmation:
+            raise serializers.ValidationError("Passwords do not match")
```
beygorghor, line 62: what about the password validation on your other PR ?

beygorghor (approved):

api/account_setup/serializers.py
tdethier: You're right, since it has now been merged, I should handle that here too!

Bewi (approved):
LGTM, just some thinking out loud: 

I wonder if we could re-use the create_snt_account for Demo and Dummy BFA.
Feels like it would prevent duplication of logic.

All the geo json logic, I thought we already had that in iaso, is it really custom for SNT ?
(Mainly thinking about the transform_geo_json_file)

tdethier:
@Bewi 

> I wonder if we could re-use the create_snt_account for Demo and Dummy BFA.
> Feels like it would prevent duplication of logic.

Good question! Maybe this could be used indeed


> All the geo json logic, I thought we already had that in iaso, is it really custom for SNT ?

No, we don't have anything on geo json, only gpkg :sweat_smile:

PR #257, opened 2026-04-10
Fix SNT-397: Impact caching

michiwend:

Add three levels of caching to scenario comparison: Backend view-level, react queries, pre-fetching

### Related JIRA tickets

SNT-397

## Changes

* React query cache until invalidation. Obvious fix, so frontend doesn't refetch all the time
* Keep old data in impact widgets to prevent visual flicker when moving the slider
* Backend view-level caching of entire impact responses by default for 7 days (configurable by account). Blazingly fast because not even serialisation is needed. Saves us a ton of requests to the impact database
* Smart-ish pre-querying of impact data for multiple years. The front-end loads the current year range and when done it pre-fetches the two years around each bound of the year range. This happens completely in the background and means that when the user moves the year slider, those responses are likely already there.
* Planned debouncing of the slider in my opinion is no longer needed, data will just show immediately as you slide

## How to test

* Create a new scenario
* Loading times for impact data is long
* Two years on each side of the range are fetched additionally
* Slide the year slider, prefetched data should appear immediately
* Reload entire page, backend responds fast with cached responses
* Manipulate scenario, backend cache is invalidated, long loading times again


js/src/domains/compareCustomize/hooks/useGetScenarioImpact.ts
```diff
@@ -29,11 +29,13 @@ export const useGetScenarioImpact = (
         queryFn: () =>
             getRequest(`/api/snt_malaria/impact/?${params.toString()}`),
         options: {
+            staleTime: Infinity,
             cacheTime: Infinity,
             enabled:
                 (enabledOverride ?? true) &&
                 Boolean(scenarioId) &&
                 Boolean(ageGroup),
+            keepPreviousData: true,
```
Bewi, line 38: I never used that keepPreviousData, what does it actually mean ?
staleTime & cacheTime are based on the key you provide I believe.
Would this one ignore it ?
Does it mean that we might have issues if we nagivate to much ?

js/src/domains/compareCustomize/hooks/usePrefetchYearsImpact.ts
```diff
@@ -0,0 +1,89 @@
...
+            !yearBounds ||
+            !selectedAgeGroup ||
+            scenarioIds.length === 0
+        )
+            return;
+
+        const [min, max] = yearBounds;
+
+        const years: [number, number][] = [];
+        for (let d = 1; d <= PREFETCH_YEARS; d++) {
```
Bewi: You decided to not use `i` just to mess up with my brain, didn't you ?

Bewi: This feels overly complicated for what it does, I had to write it down :D 

So let's say I have:
`yearFrom=2025.  and yearTo=2030`.

PREFETCH_YEAR is 2. So we go with two iteration of 1 and 2.

You create an array of:
`[2026, 2030], [2024, 2030], [2025, 2031], [2025,2029]` to then check which one you want to keep fitting within the bound.

Then the same with 2 years spread. 

Wouldn't it be clearer to have to two lists and go through ?

yearsFrom = [2024, 2025, 2026, 2027, 2028]
yearsTo = [2028, 2029, 2030, 2031, 2032]

Not sure it'll be clearer, but you get my point :D 
If you don't want this approach, I'd request at least a small comment to make it easier when we come in here.

```
lowerStartYearRange = yearFrom - PREFETCH_YEARS
higherEndYearRange = yearTo + PREFETCH_YEARS
yearRange = PREFETCH_YEARS + 1;
for (let i = 0; i < yearRange; i++) {
   startYear = lowerStartYearRange + i;
   if (startYEar < bound) continue;
   for let(j = 0; j < yearRange; j++) {
      const endYear = higherEndYearRange - j;
      if (endYear > bound) continue;
      years.push([startYear, endYear]
  }
}
```

Bewi: Or split this in different functions :D

api/impact/cache.py
```diff
@@ -0,0 +1,54 @@
...
+def _scenario_stamp(scenario_id) -> str | None:
+    row = Scenario.objects.filter(pk=scenario_id).values_list("updated_at", flat=True).first()
+    if row is None:
+        return None
+    return str(int(row.timestamp()))
+
+
+def _build_key(scenario_id, age_group, year_from, year_to) -> str | None:
+    stamp = _scenario_stamp(scenario_id)
+    if stamp is None:
```
Bewi: could be a 1 liner I think: 
`return "whatever" if stamp`

Bewi (changes requested):
Looks ok, only a few stuff I would try to either document or make it easier to read for the one coming after.

Also, I am missing some tests for the cache at least.

js/src/domains/compareCustomize/hooks/usePrefetchYearsImpact.ts
michiwend: You are right! I arrived there iterating from a previous approach so it became a bit unreadable 🫠

js/src/domains/compareCustomize/hooks/useGetScenarioImpact.ts
michiwend: it means that if there was already data in the widget, it stays until new data arrives. It prevents visual flicker when you move the slider

michiwend: > Would this one ignore it ?

No, those still work exactly the same, per query key.


> Does it mean that we might have issues if we nagivate to much?

No. It only ever holds onto the data from the previous query key

Bewi (approved):
LGTM

PR #258, opened 2026-04-13
SNT-418 turn the cost settings page into a proper interventions management screen

Bewi:

Add intervention page to modify intervention target population and cost settings

### Related JIRA tickets

SNT-418

## Changes

- Remove settings page (which only had cost settings)
- A new page to list interventions
- Add the target pop to an intervention

## How to test

Go to SNT Malaria (follow setup from the readme if you don't have any account / user yet).
Open the side nav, you should have an intervention entry and no settings anymore.
Going to intervention should show you list of interventions grouped per category.
First intervention should be selected by default. 
Form should be displayed and prefilled with intervention values.
Modifying those values and saving should properly persist the intervention.

Few more advanced testing/notes: 

- The population currently uses metric layers, which is not ideal but we need to define a way to specify which one is for population.
- There are few intervention category which requires two population inputs: ITN Routine, PMC & SMC.
   For those, you can try to modify one, save, modify the second one, save, hard refresh, modify again and make sure selection is not cleared out.

## Print screen / video

image

image

## Notes

N/A

## Doc

N/A


api/interventions/serializers.py
```diff
@@ -21,3 +23,50 @@ class Meta:
...
+            "target_population",
+            "cost_breakdown_lines",
+        ]
+
+    def update(self, instance, validated_data):
+        cost_breakdown_lines_data = validated_data.pop("cost_breakdown_lines", [])
+        intervention = super().update(instance, validated_data)
+
+        # Delete existing cost breakdown lines and create new ones based on the provided data
+        with transaction.atomic():
```
michiwend, line 63: Should the transaction not include the `super.update()`? If a cost-line creation fails, the intervention fields are already committed.

api/interventions/serializers.py
```diff
@@ -21,3 +23,50 @@ class Meta:
...
+            instance.cost_breakdown_lines.all().delete()
+            for line_data in cost_breakdown_lines_data if cost_breakdown_lines_data else []:
```
michiwend: knit-pick: `cost_breakdown_lines_data` is already defaulted to `[]` via `.pop("cost_breakdown_lines", [])`.

api/interventions/serializers.py
```diff
@@ -21,3 +23,50 @@ class Meta:
...
+                InterventionCostBreakdownLineSerializer.create(
```
michiwend: You could use `InterventionCostBreakdownLineSerializer(data=...).save()` or just call the model manager directly `InterventionCostBreakdownLine.objects.create(...)`

api/interventions/views.py
```diff
@@ -1,13 +1,34 @@
...
+    InterventionSerializer,
+)
 from plugins.snt_malaria.models import Intervention
 
 
 class InterventionViewSet(viewsets.ModelViewSet):
     serializer_class = InterventionSerializer
     ordering_fields = ["id", "name"]
-    http_method_names = ["get", "options"]
+    http_method_names = ["get", "options", "put"]
```
michiwend, line 17: Adding "put" globally means the ModelViewSet's default update action now accepts `PUT` requests at `/api/snt_malaria/interventions/<id>/` using the base `InterventionSerializer`.

api/interventions/views.py
```diff
@@ -1,13 +1,34 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    @action(detail=True, methods=["put"], serializer_class=InterventionDetailWriteSerializer)
+    def update_details(self, request, pk):
+        intervention = self.get_object()
+        serializer = self.get_serializer(intervention, data=request.data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
michiwend: `serializer` in `Response(serializer.data, ...` is the write serializer (line 27). If you want to serialize the updated interventions you need to use `InterventionDetailSerializer(intervention).data`

api/interventions/views.py
```diff
@@ -1,13 +1,34 @@
...
         return Intervention.objects.filter(intervention_category__account=self.request.user.iaso_profile.account)
+
+    @action(detail=True, methods=["get"], serializer_class=InterventionDetailSerializer)
+    def details(self, request, pk):
+        result = self.get_object()
+        serializer = self.get_serializer(result)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    @action(detail=True, methods=["put"], serializer_class=InterventionDetailWriteSerializer)
+    def update_details(self, request, pk):
```
michiwend, line 30: `update_details` can be called by any (authenticated) user with account access; there's no `SETTINGS_WRITE` permission check on the viewset or the action, only a `SETTINGS_READ` guard at the route level. The deleted InterventionRow used it to hide the edit button from users without `SETTINGS_WRITE` permission

tests/api/interventions/test_serializers.py
```diff
@@ -0,0 +1,88 @@
+from iaso.models.base import Account
+from plugins.snt_malaria.api.interventions.views import InterventionDetailWriteSerializer
```
michiwend: should be `from plugins.snt_malaria.api.interventions.serializers import InterventionDetailWriteSerializer`

js/src/domains/interventions/components/InterventionFormWrapper.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ) {
+            let defaultPop = defaultTargetPopulationSettings[
+                interventionDetails.code
+            ] || [''];
+
+            defaultPop = defaultPop.map(pop =>
+                metricTypes.find(metric => metric.code === pop) ? pop : '',
+            );
+
+            interventionDetails.target_population = defaultPop; // Set default target population if not already set
```
michiwend: not sure if we should mutate the `interventionDetails` as it is a ref to a query cached object. Maybe something like this:
```ts
const details = { ...interventionDetails, target_population: defaultPop };
formik.setValues(details);
```

js/src/domains/interventions/hooks/useGetInterventionDetails.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useSnackQuery } from 'Iaso/libs/apiHooks';
+import { InterventionDetails } from '../../planning/types/interventions';
+
+export const useGetInterventionDetails = ({
+    interventionId,
+}: {
+    interventionId?: number;
+}): UseQueryResult<InterventionDetails, Error> => {
+    return useSnackQuery({
+        queryKey: [`interventionDetails_${interventionId}`],
```
michiwend: Knit-pick: Using a structured array key like `['interventionDetails', interventionId]` is more idiomatic and makes it easier to use `queryClient.invalidateQueries` with partial key matching

js/src/domains/interventions/components/InterventionFormWrapper.tsx
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const defaultTargetPopulationSettings: Record<string, string[]> = {
+    iptp: ['pop_pw'],
+    itn_campaign: ['pop_total'],
+    itn_routine: ['pop_0_5', 'pop_pw'],
+    pmc: ['pop_0_1', 'pop_1_2'],
+    smc: ['pop_0_5'],
+    vacc: ['pop_vaccine_5_36_months'],
+};
```
michiwend, line 29: we also have 
```ts
const interventionPopulationSettings: Record<string, string[]> = {
    itn_routine: ['pop_0_5', 'pop_pw'],
    pmc: ['pop_0_1', 'pop_1_2'],
};
```
in `TargetPopulationForm.tsx`. Can they be consolidated?

js/src/domains/interventions/components/InterventionForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+                            }
+                            getErrors={field => getChildError(field, index)}
+                        />
+                    ))}
+                <Button
+                    variant="text"
+                    sx={{ alignSelf: 'flex-start' }}
+                    onClick={() =>
+                        addChildValue('cost_breakdown_lines', { year: 2024 })
+                    }
```
michiwend, line 121: Hardcoded year, is that intended?

js/src/domains/interventions/components/InterventionFormWrapper.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const InterventionFormWrapper: FC<Props> = ({ interventionId }) => {
```
michiwend, line 31: Consider adding loading animations when changing the interventions and when saving an intervention. E.g. disable the button or something.

michiwend (changes requested):
Nicely build! So much better to adjust all the settings in one place 👍 
I found a few things, have a look.

One knit about the UI, can we remove the scrollbars and have some more space to the header (see screenshot)

Image

api/interventions/serializers.py
Bewi: Good question, I couldn't decide to be honnest, should be block entire update or not, I think I'll block all so feedback to the user will make more sense

js/src/domains/interventions/components/InterventionForm.tsx
Bewi: Yes, we got another story to remove the year

api/interventions/serializers.py
Bewi: I wish I wouldn't have to do this to be honnest as we already have the field definition.
But ok, I'll still use the Serializer so I can verify if it is valid.

api/interventions/views.py
Bewi: Not sure what is your point here, I don't see the problem.

api/interventions/views.py
Bewi: Added permissions and tests

js/src/domains/interventions/components/InterventionFormWrapper.tsx
Bewi: I know it is weird, but they don't mean the same thing. 
The one from TargetPopulationForm is more there to describe inputs available.
The one from InterventionFormWrapper is to define a default value on those inputs :D

api/interventions/views.py
michiwend: The point is, that your new action creates an explicit PUT route for `/api/snt_malaria/interventions/{pk}/update_details/` so you don't need the `PUT` verb in the array because that creates an addiional default put endpoint. Not dramatic, just not needed.

michiwend (approved):
I still hate the scrollbar :D but otherwise LGTM. The think about the PUT request I commented, you can decide if you want to fix that or not.

PR #259, opened 2026-04-15
Feature SNT-339: comparative averted cases

michiwend:

Remove averted cases from backend/API, calculate comparative averted cases in frontend

### Related JIRA tickets

SNT-339

## Changes

The "Cost per averted case" widget in Compare & Customize now calculates the metric comparatively against the selected baseline scenario, rather than calculating averted cases against the population within a scenario in the backend.

#### Backend

* Removed `averted_cases`, `cost`, and `cost_per_averted_case` from the impact API response (service, serializers, and tests). These are no longer computed server-side.

#### Cost per averted case chart

* Calculates cost per averted case on the frontend by comparing each scenario's budget and uncomplicated cases against the baseline: `(scenario.cost - baseline.cost) / (baseline.cases - scenario.cases)`
* Switched from vertical to horizontal bar chart with a yellow baseline reference line at zero, labelled "Baseline"
* Confidence intervals shown as error bars
* Scenarios that don't meaningfully avert cases (< 1) are filtered out with a dedicated info message
* New tooltip shows cost per averted case, relative cost, and cases averted
* New ChartTooltip component with compact, consistent styling (matching map tooltips) used across all three chart widgets
* Added optional help icon tooltip to Card component, used for explaining cost per averted case
* Fixed formatBigNumber to correctly handle negative values and apply .toFixed(2) consistently

At least two scenarios need to be selected
image

No cases are averted
image

Different positive/negative bar constellations
image
image
image

## How to test
* Prepare three scenarios with valid interventions that match impact data and generate budgets for them. Ideally they yield significantly differetnt cases and costs
* Prepare one scenario without cost
* Go to Compare and Customize, you should see "select at least two scenarios" message in the widget
* Select two scenarios, one of which has no budget data, you should see the "no budget data" message
* Select two scenarios with budget data, select one with **less or equal** cases as your baseline. You shoud see a "no cases averted" message in the widget
* Select two scenarios with budget data, select the one with the highest cases as your baseline. You should see averted cases bars
* Select three scenarios with budget data, you should see two bars

### DIFF
--- a/js/src/domains/compareCustomize/components/charts/ChartTooltip.tsx
+++ b/js/src/domains/compareCustomize/components/charts/ChartTooltip.tsx
@@ -1,0 +1,65 @@
+import React, { FC, ReactNode } from 'react';
+import { Box } from '@mui/material';
+import { SxStyles } from 'Iaso/types/general';
+
+export type ChartTooltipRow = {
+    label: string;
+    value: string;
+    color?: string;
+};
+
+type Props = {
+    title: ReactNode;
+    rows: ChartTooltipRow[];
+};
+
+const styles = {
+    root: {
+        backgroundColor: 'common.white',
+        border: '1px solid',
+        borderColor: 'grey.300',
+        borderRadius: 0.5,
+        boxShadow: '0 1px 3px rgba(0,0,0,0.15)',
+        px: 1,
+        py: 0.75,
+        fontSize: '0.75rem',
+        lineHeight: 1.4,
+    },
+    title: {
+        fontSize: '0.75rem',
+        display: 'block',
+        mb: 0.5,
+    },
+    row: {
+        display: 'flex',
+        alignItems: 'center',
+        gap: 0.75,
+    },
+    dot: {
+        width: 8,
+        height: 8,
+        borderRadius: '50%',
+        flexShrink: 0,
+    },
+} satisfies SxStyles;
+
+export const ChartTooltip: FC<Props> = ({ title, rows }) => (
+    <Box sx={styles.root}>
+        <Box component="b" sx={styles.title}>
+            {title}
+        </Box>
+        {rows.map(row => (
+            <Box key={row.label} sx={styles.row}>
+                {row.color && (
+                    <Box sx={{ ...styles.dot, backgroundColor: row.color }} />
+                )}
+                <Box component="span" sx={{ color: 'text.secondary' }}>
+                    {row.label}:
+                </Box>
+                <Box component="span" sx={{ fontWeight: 700 }}>
+                    {row.value}
+                </Box>
+            </Box>
+        ))}
+    </Box>
+);
--- a/js/src/domains/compareCustomize/components/charts/ChartTooltip.tsx
+++ b/js/src/domains/compareCustomize/components/charts/ChartTooltip.tsx
@@ -1,5 +1,5 @@
 import React, { FC, ReactNode } from 'react';
-import { Box } from '@mui/material';
+import { Box, Typography } from '@mui/material';
 import { SxStyles } from 'Iaso/types/general';
 
 export type ChartTooltipRow = {
@@ -27,6 +27,7 @@ const styles = {
     },
     title: {
         fontSize: '0.75rem',
+        fontWeight: 700,
         display: 'block',
         mb: 0.5,
     },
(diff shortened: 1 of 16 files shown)


js/src/domains/compareCustomize/components/charts/ChartTooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Box sx={styles.root}>
+        <Box component="b" sx={styles.title}>
+            {title}
+        </Box>
+        {rows.map(row => (
+            <Box key={row.label} sx={styles.row}>
+                {row.color && (
+                    <Box sx={{ ...styles.dot, backgroundColor: row.color }} />
+                )}
+                <Box component="span" sx={{ color: 'text.secondary' }}>
```
Bewi: Any reason for using Box and not a Typography or something more descriptive ?

js/src/domains/compareCustomize/components/charts/CostPerAvertedCaseCard.tsx
```diff
@@ -8,94 +8,248 @@ import {
...
     new Intl.NumberFormat(undefined, {
         minimumFractionDigits: 0,
         maximumFractionDigits: 2,
     }).format(value);
 
+/**
+ * D3-style "nice" step size for axis ticks (see d3-scale tickStep).
+ * Thresholds ≈ √50, √10, √2 are geometric means between 10, 5, 2, 1.
+ */
+const niceStep = (range: number, count: number): number => {
```
Bewi: I wonder if we should have those two helper in a libs/utils for charts

js/src/domains/compareCustomize/components/charts/CostPerAvertedCaseCard.tsx
```diff
@@ -8,94 +8,248 @@ import {
...
+            baselineScenarioId,
+        ],
     );
 
+    const { domain, ticks } = useMemo(
+        () => computeNiceTicks(chartData),
+        [chartData],
+    );
+
+    let emptyMessage: string;
```
Bewi: useMemo ?

js/src/domains/compareCustomize/utils/chartData.ts
```diff
@@ -53,41 +55,112 @@ export const buildPrevalenceChartData = (
...
+ *
+ * Only scenarios that meaningfully avert cases (≥ 1) are included.
+ *
+ * Only uncomplicated cases (excluding severe) are used for averted cases.
+ * Confidence intervals use the worst/best case averted bounds
+ * (baseline.lower − scenario.upper and baseline.upper − scenario.lower).
+ * The baseline itself is never shown (it has no comparison target).
  */
-export const buildCostChartData = (
+export const buildCostPerAvertedCaseChartData = (
```
Bewi, line 97: To be honnest, I don't understand a thing of what this actually calculate the cost per averted case :D

js/src/domains/compareCustomize/utils/chartData.ts
```diff
@@ -53,41 +55,112 @@ export const buildPrevalenceChartData = (
...
+    baselineScenarioId: number | undefined,
+): CostPerAvertedCaseResult => {
+    const empty: CostPerAvertedCaseResult = {
+        data: [],
+        hasInsufficientAverted: false,
+    };
+    if (baselineScenarioId === undefined) return empty;
+
+    const baselineImpact = impactsByScenarioId.get(baselineScenarioId);
+    const b = baselineImpact?.number_cases;
```
Bewi: can we name this a bit more explicitly ? 
baseLineCases or avertedCases or smth that makes it easier to understand what we do later ?

js/src/domains/compareCustomize/utils/chartData.ts
```diff
@@ -53,41 +55,112 @@ export const buildPrevalenceChartData = (
...
+
+    let hasInsufficientAverted = false;
+
+    const data = scenarios
+        .filter(s => s.id !== baselineScenarioId)
         .map((scenario): CostPerAvertedCaseDatum | null => {
             const impact = impactsByScenarioId.get(scenario.id);
-            const metric = impact?.cost_per_averted_case;
-            const costValue = metric?.value;
+            const s = impact?.number_cases;
```
Bewi: same here, I mean, s is a beaufitully shaped letter but ... hard to remember what it is later on

Bewi (changes requested):
Overall it works and look ok, I would just like to make it a bit clearer :)

Bewi (approved):
LGTM

PR #260, opened 2026-04-16
SNT-396: Added an isScenarioEditable which check on lock and user perms

Bewi:

Don't allow rule reordering and edition if a scenario is locked

### Related JIRA tickets

SNT-396

## Changes

Added a calculated field on the contextto know if the scenario is locked and if user has auth: isScenarioEditable.
Which lives aside of canEditScenario (which only let us know if a user has access).
This is because we want to hide the lock / unlock only if user does not have access. If the scenario is locked, we still want to see this action.

Replaced canEditScenario with isScenarioIsEditable everywhere except for the lock button display.

## How to test

Go to SNT, open a scenario and lock it.
You should not be able to edit a rule or to drag and drop.
You should be able to unlock it.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/tests/api/scenario_rules/test_views.py
+++ b/tests/api/scenario_rules/test_views.py
@@ -42,6 +42,15 @@ class ScenarioRuleAPITestCase(ScenarioRulesTestBase):
         response = self.client.get(self.BASE_URL, {"scenario": self.scenario.id})
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
+    def test_list_locked_scenario_rules(self):
+        self.scenario.is_locked = True
+        self.scenario.save()
+
+        self.client.force_authenticate(user=self.user_with_full_perm)
+        response = self.client.get(self.BASE_URL, {"scenario": self.scenario.id})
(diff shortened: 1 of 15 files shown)


tests/api/scenario_rules/test_views.py
```diff
@@ -42,6 +42,15 @@ def test_list_scenario_rules_unauthenticated(self):
         response = self.client.get(self.BASE_URL, {"scenario": self.scenario.id})
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
+    def test_list_locked_scenario_rules(self):
+        self.scenario.is_locked = True
+        self.scenario.save()
```
tdethier: You forgot to use your utility method in this file 😅

tdethier (approved):
Alright, rules cannot be edited when a scenario is locked, good job 👍 

At the moment, there's nothing in the backend that prevents from editing a scenario (name, years...) or reordering its rules if it's locked. I would probably fix this here as well, but if you want to do this in another ticket/PR, it's up to you

PR #263, opened 2026-04-16
fix SNT-416: fetch all org units for dropdowns when zoomed

michiwend:

Fix rule drop downs showing error when selected org unit is not in current zoomed region

### Related JIRA tickets

SNT-416

## Changes

Before: The rule form dropdowns shared the map's org units request. Filtering by region also reduced the dropdown options, breaking selections from other regions.

Now: The dropdowns fetch the full unfiltered list independently. This uses the same query key as the map's "All" request, so react query's cache is shared - no extra network call when the page starts unfiltered. When the page starts filtered, a brief loading state appears on the dropdowns while they fetch.

Fetching org units with geometry in the dropdown request seems overkill but it allows to share the query key with the map's, enabling cache sharing.

Also refactored `useGetOrgUnitsByType` into `useGetOrgUnits` with an options object (`orgUnitParentId`, `orgUnitTypeId`, `withGeometry`), and added `orgUnitTypeId` filtering to the map request to reduce response size.

Update:
Also the Compare & Customize page uses the same request now so it also uses the same cache and doesn't re-fetch.

## How to test

Prepare a scenario with a rule that has a few org units in the included dropdown

Scenario 1, start unfiltered:
* Load rule builder from scratch (unfiltered), observe the org unit requests (one for the region dropdown, one for all intervention org units)
* Edit the rule, no additional requets should appear
* Change region filter to different regions, one request per selection. The existing dropdown selections stay stable, no error

Scenario 2, start filtered:
* Load rule builder with a region filter (by using deep link) observe the org unit requests (one for the region dropdown, one for the filtered intervention org units)
* Edit the rule, you should see spinners in the dropdowns and an additional request of all org units, selections eventuall appear
* Change region filter to different regions, one request per selection. The existing dropdown selections stay stable, no error

Initial state is all, drop downs share query key with the initial org unit fetch, no further fetch

https://github.com/user-attachments/assets/e5bd2e2f-4be1-4d07-8c2b-19046e170afc

Initial state is zoomed, only partial org units are fetched, drop downs trigger full fetch

https://github.com/user-attachments/assets/f6567cdf-1c6b-4ce5-b20c-ebc89a48d98e

### DIFF
--- a/js/src/components/OrgUnitSelect.tsx
+++ b/js/src/components/OrgUnitSelect.tsx
@@ -29,10 +29,12 @@ export const OrgUnitSelect: FC<Props> = ({
 
     const { data: accountSettings } = useGetAccountSettings();
 
+    const focusTypeId = accountSettings?.focus_org_unit_type_id;
     const { data: orgUnitsByType, isLoading: isLoadingOrgUnits } =
         useGetOrgUnits({
-            orgUnitTypeId: accountSettings?.focus_org_unit_type_id,
+            orgUnitTypeId: focusTypeId,
             withGeometry: false,
+            enabled: !!focusTypeId,
         });
 
     const handleOrgUnitChange = (e: SelectChangeEvent<number>) => {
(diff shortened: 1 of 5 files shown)


Bewi (approved):
LGTM

PR #264, opened 2026-04-16
Feature SNT-415: region filter data layers

michiwend:

Add the same region filter to _Data Layers_ that we use in the planner

### Related JIRA tickets

SNT-415

## Changes

In addition the requests are also scoped by intervention org unit type

## Print screen / video

https://github.com/user-attachments/assets/51d9ffeb-f253-481e-9366-ddfdbd653473

### DIFF
--- a/js/src/domains/dataLayers/index.tsx
+++ b/js/src/domains/dataLayers/index.tsx
@@ -1,10 +1,12 @@
 import React, { FC, useCallback, useState } from 'react';
-import { Card, Typography } from '@mui/material';
-import { LoadingSpinner, useSafeIntl } from 'bluesquare-components';
+import { Card, Stack, Typography } from '@mui/material';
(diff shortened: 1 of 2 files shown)


js/src/domains/dataLayers/index.tsx
```diff
@@ -33,12 +37,27 @@ const styles = {
     },
 } satisfies SxStyles;
 
+type DataLayersParams = {
+    displayOrgUnitId?: number;
+};
+
 export const DataLayers: FC = () => {
     const { formatMessage } = useSafeIntl();
+    const { displayOrgUnitId } = useParamsObject(
```
Bewi, line 46: I wonder if it would be a good idea to move this and the redirect to the OrgUnitSelect with a flag: redirectOnChange

Bewi (approved):
LGTM

PR #265, opened 2026-04-20
SNT-196 correctly handle the years in the budgeting cost dataframe

Bewi:

Remove the year from `cost breakdown lines` and only keep rows for the first year found.

### Related JIRA tickets

SNT-196

## Changes

- Removed `year` from model, views, filters and serializers
- Removed `year` when seeding interventions
- Removed hardcoded `year` on form and properly reset state when changing intervention
- Align tests

## How to test

Before pulling, add some costs lines from the interventions page and go to admin to change the year to 2025 and later.
Run migrations and verify that it is gone properly.
Try to edit interventions costs and verify that costs are properly modified.

To test further, add some costs to `CM`, make sure to select `Other` as unit.
And select total population as target population.
Do the same for `IPTP`.

Go to a scenario, create a rule for `CM` and another one for `IPTP`

Go to the budget tab and try to run the budget, it should work.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


tests/api/intervention_cost_breakdown_lines/test_views.py
michiwend: no longer needed

michiwend (approved):
Overall LGTM,

two things:

`year` is not needed here anymore:
https://github.com/BLSQ/snt-malaria/blob/74a5d2bc60ac2e37219c1de982571c010ac710a8/js/src/domains/planning/types/interventions.ts#L82-L92

and the other one see comment

PR #266, opened 2026-04-20
SNT-431: Add rules to the demo account

Bewi:

Create two scenarios with rules when recreating the demo account

### Related JIRA tickets

SNT-431

## Changes

Generate rules when creating the demo account and don't calculate fake budget anymore.

## How to test

On snt, run `manage.py recreate_demo_account`
Login using provided credentials on command log and verify that there are 2 scenarios.
Each of them work (you can check the rules and such) and run the budget, it should work as well.

## Print screen / video

image

image

image

## Doc

### DIFF
--- a/management/commands/support/demo_scenario_seeder.py
+++ b/management/commands/support/demo_scenario_seeder.py
@@ -2,23 +2,21 @@
 Create a demo scenario with intervention assignments for Burkina Faso
 """
 
-import random
-
 from datetime import date
 
 from snt_malaria_budgeting import DEFAULT_COST_ASSUMPTIONS, BudgetCalculator
 
 from iaso.models import OrgUnitType, User
 from iaso.models.data_store import JsonDataStore
-from iaso.models.metric import MetricValue
+from iaso.models.metric import MetricType
 from plugins.snt_malaria.api.budget.utils import (
     build_cost_dataframe,
     build_interventions_input,
     build_population_dataframe,
 )
 from plugins.snt_malaria.models.budget import Budget
-from plugins.snt_malaria.models.intervention import Intervention, InterventionAssignment
-from plugins.snt_malaria.models.scenario import Scenario
+from plugins.snt_malaria.models.intervention import Intervention
+from plugins.snt_malaria.models.scenario import Scenario, ScenarioRule, ScenarioRuleInterventionProperties
 
 
 class DemoScenarioSeeder:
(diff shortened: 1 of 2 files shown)


management/commands/support/demo_scenario_seeder.py
```diff
@@ -234,3 +148,123 @@ def _create_budget_for_scenario(self, scenario, created_by):
...
+            priority=1,
+            color="#26C6DA",
+            matching_criteria={"all": True},
+        )
+
+        # Only required for json matching criteria as it uses matching_orgunits field.
+
+        matching_criteria_seasonal = {
+            "and": [
+                {">=": [{"var": seasonality_precipitation_metric_type.id}, "seasonal"]},
```
michiwend: works but should be `=` for ordinal metrics

management/commands/support/demo_scenario_seeder.py
```diff
@@ -234,3 +148,123 @@ def _create_budget_for_scenario(self, scenario, created_by):
...
+            coverage=1,
+        )
+        ScenarioRuleInterventionProperties.objects.create(
+            scenario_rule=itn_dual_ai_rule,
+            intervention=interventions.filter(code="itn_routine", name="Dual AI").first(),
+            coverage=1,
+        )
+        ScenarioRuleInterventionProperties.objects.create(
+            scenario_rule=itn_pbo_rule,
+            intervention=interventions.filter(code="itn_routine", name="PBO").first(),
```
michiwend: `.filter().first()` is a bit fragile. In case those interventions ever change, this would just return `None`. Can we not use `.get()` here, too?

management/commands/support/demo_scenario_seeder.py
```diff
@@ -234,3 +148,123 @@ def _create_budget_for_scenario(self, scenario, created_by):
         )
 
         return budget
+
+    def _create_scenarion_(
```
michiwend: `_create_scenario` ?

michiwend (approved):
Tested and looks overall good, few knit picks in the comments

management/commands/support/demo_scenario_seeder.py
Bewi: Nice catch, forgot to change that one

management/commands/support/demo_scenario_seeder.py
Bewi: Yep, leftover, thanks for the catch

PR #267, opened 2026-04-21
Feature SNT-407: Fake impact provider

michiwend:

Add a fake impact provider for demo purposes

### Related JIRA tickets

SNT-407

## Changes

Adds a new fake impact provider that generates epidemiologically _plausible_ malaria impact metrics on the fly, without requiring any external data source. Uses population data from the data-layers as the only real input, and five generic intervention reference codes (`impact_1` through `impact_5`) that an account's real interventions can be mapped to via `Intervention.impact_ref`.

The five reference codes represent intervention "strength tiers": `impact_1` is the weakest (modest reduction, cases slowly grow over time) and `impact_5` is the strongest (large initial reduction plus sustained decline). Real effect sizes are loosely grounded in 2024-2026 literature (ITN ~13%, ITN+IRS combined ~50%, etc.) but are primarily chosen to produce visually meaningful differences between scenarios in the Compare & Customize UI.

For the fake provider to work, you need to configure it for a given account using the _Impact provider configuration_
in Django admin (see screenthot).

As provider choose `Fake (demo)`

The configuration requires the `population_metric_code` to point to the respective metric type's code for the total population data of the account:
```json
{
  "population_metric_code": "POPULATION"
}
```

`Cache ttl seconds` can be set to 0 as there is no need for caching.

After setup, the account's interventions need to be mapped to any of the 5 impact "strength tiers" using the impact provider reference field. So, for example, map CM to `impact_1` to get the lowest impact and map a bed net intervention to `impact_5` to get the highest impact.

Additional changes in the PR are
* New `provider_meta` field in the impact response
* Front-end renders info box when `provider_key == "fake"` in the provider_meta

## How to test

* Set up the provider as described
* generate scenarios with the interventions that you previously mapped
* Compare & Customize should show impact for your scenario for two age groups (0-5, 0-100)
* You should see an info box about the fake data
* A scenario with `impact_5` interventions should outperform a scenario with `impact_1` interventions (lower cases, better prevalence reduction)

## Print screen / video

Compare & Customize for Burkina Faso and the fake provider
image

Django admin configuration
image

Intervention mapping (CM to impact_1)
image


js/src/domains/compareCustomize/components/ConfigurationPanel.tsx
```diff
@@ -165,6 +159,17 @@ const styles = {
     },
 } satisfies SxStyles;
 
+const alertBoxSx = (
```
Bewi: This has only two options, feels like it'll be computed everytime you call it. 
My suggestion would be to have three styles: 
```
const styles = {
    alertBox: {...},
    alertBoxWarning: {backgroundColor , "& .MuiAlert-icon"},
    alertBoxInfo: {backgroundColor, "& .MuiAlert-icon"}
}
```

Then in sx bellow, you can use an array `sx={[styles.alertBox, styles.alertBoxWarning]}`

Official doc:
https://mui.com/system/getting-started/the-sx-prop/#array-values

js/src/domains/compareCustomize/components/ConfigurationPanel.tsx
```diff
@@ -351,88 +358,94 @@ export const ConfigurationPanel: FC<Props> = ({
...
+                        {formatMessage(MESSAGES.orgUnitsNotFound)}
+                        {showNotFoundDetails && (
+                            <Box sx={styles.chipContainer}>
+                                {orgUnitsNotFound.flatMap(({ scenario, orgUnits }) =>
+                                    orgUnits.map(ou => (
+                                        <Chip
+                                            key={`${scenario.id}-${ou.org_unit_id}`}
+                                            label={ou.org_unit_name}
+                                            size="small"
+                                            sx={{
```
Bewi: Move this to the styles.
Could the white use a theme color ?

Bewi (approved):
LGTM
If you have time to just fix the style thing would be nice

PR #269, opened 2026-04-22
Fix/snt 385 idm age group under5

michiwend:

Fix incidence scaling for under5 age group in IDM provider

### Related JIRA tickets

SNT-385

## Changes

Add the `population_u5` column to the `admin_info` model after it was added by IDM to their dataset. Uses `population_u5` for scaling incidence to cases for this age group.

Additionally this PR adds the openhexa pipeline that is used to create and load the IDM impact data.


Bewi (approved):
LGTM

PR #271, opened 2026-04-23
SNT-437 scenario form year range slider

Bewi:

Change year input when creating / editing a scenario to a slider

### Related JIRA tickets

SNT-437

## Changes

- Moved slider from impact to a component
- Use the slider in both impact and scenario form

## How to test

Go to SNT, create a scenario, play with the slider. 

Edit the scenario, play with the slider.

## Print screen / video

image

## Notes

N/A

## Doc

N/A

### DIFF
--- a/js/src/domains/scenarios/hooks/useScenarioFormState.tsx
+++ b/js/src/domains/scenarios/hooks/useScenarioFormState.tsx
@@ -9,8 +9,8 @@ export type ScenarioFormValues = {
     id?: number;
     name: string;
     description: string;
-    start_year?: number;
-    end_year?: number;
+    start_year: number;
+    end_year: number;
 };
 
 export const SCENARIO_YEAR_RANGE = {
@@ -82,7 +82,7 @@ const useValidation = () => {
 
 export const useScenarioFormState = (
     scenario: Scenario | undefined,
-    onSubmit,
+    onSubmit: (values: ScenarioFormValues) => void,
 ) => {
     const validationSchema = useValidation();
     const values = useMemo(() => {
--- a/js/src/domains/scenarios/hooks/useScenarioFormState.tsx
+++ b/js/src/domains/scenarios/hooks/useScenarioFormState.tsx
@@ -13,12 +13,17 @@ export type ScenarioFormValues = {
     end_year?: number;
 };
 
+export const SCENARIO_YEAR_RANGE = {
+    min: 2024,
+    max: 2035,
+} as const;
+
 const initialValues: ScenarioFormValues = {
     id: undefined,
     name: '',
     description: '',
-    start_year: undefined,
-    end_year: undefined,
+    start_year: SCENARIO_YEAR_RANGE.min,
+    end_year: SCENARIO_YEAR_RANGE.max,
 };
 
 const useValidation = () => {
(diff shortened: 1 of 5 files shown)


js/src/domains/scenarios/hooks/useScenarioFormState.tsx
```diff
@@ -13,12 +13,17 @@ export type ScenarioFormValues = {
     end_year?: number;
 };
 
+export const SCENARIO_YEAR_RANGE = {
+    min: 2024,
+    max: 2035,
+} as const;
```
michiwend, line 19: Two things:

We define this range now in the front-end and the back-end:
https://github.com/BLSQ/snt-malaria/blob/develop/api/scenarios/serializers.py#L41-L42

Maybe these could be exposed by the backend and then the second point is whether we want make them configurable or generate them based on current year?

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
michiwend, line 98: A bit picky and pre-existing but now with the min/max range of the slider this might become a problem when editing an old scenario that falls out of the current range. It would just pass the scenario's years but then the form would limit the slider to the current min/max range. So lets say we have an old scenario with 2020 but the new min is 2024. The form would show 2024 as min but the underlying value would be 2020 (see screenshot)

Image

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
```diff
@@ -13,12 +13,17 @@ export type ScenarioFormValues = {
     end_year?: number;
```
michiwend: you can drop the ? now that we don't have the inputs that allowed blank and then remove null checks here https://github.com/BLSQ/snt-malaria/pull/271/changes#diff-ac48c305df74441b90f406dee5ce6ee83ea260dda6c681dc435c899526e4856cR45-R48

michiwend (changes requested):
Tested locally, its so much easier to use than before. I found a few things worth checking:

These are not used anymore:
https://github.com/BLSQ/snt-malaria/blob/develop/js/src/domains/messages.ts#L788-L791
https://github.com/BLSQ/snt-malaria/blob/develop/js/src/domains/messages.ts#L364-L372

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
Bewi: Range was always this, not sure how you got a scenario with a start date lower than 2024

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
Bewi: We can discuss about it but I won't fix it now, this was just to change the input.
To be honnest, I don't like the min max, but it requires some thinking and validation from Fred

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
Bewi: Cleaned it up

js/src/domains/scenarios/hooks/useScenarioFormState.tsx
michiwend: I did that in django admin just to test this edge case. I don't think it is very critical but it might fall on our feat later if we allow the min/max range to change

michiwend (approved):

PR #272, opened 2026-04-23
fix(rules): consistent matching pipeline for match-all and criteria

michiwend:

Consistent matching pipeline for match-all and criteria-based rules

## Changes

Now both, criteria-based and match-all rules generate `org_units_matched` which caches them on the rule.

In addition we re-match org units when updated, regardless of whether the matching criteria were updated. This helps with avoiding stale org units in old rules

### DIFF
--- a/models/scenario.py
+++ b/models/scenario.py
@@ -14,7 +14,7 @@ from iaso.utils.models.soft_deletable import (
     SoftDeletableModel,
 )
 from iaso.utils.validators import JSONSchemaValidator
-from plugins.snt_malaria.models.account_settings import get_intervention_org_unit_type_id, get_intervention_org_units
+from plugins.snt_malaria.models.account_settings import get_intervention_org_units
 
 
 class Scenario(SoftDeletableModel):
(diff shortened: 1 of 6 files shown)


models/scenario.py
```diff
@@ -180,30 +180,31 @@ def resolve_matched_org_units(account, matching_criteria):
...
-        - criteria: evaluates against MetricValues, also scoped to the intervention level.
+
+        1. start from the account's valid geo-located org units at the configured
+           intervention_org_unit_type level (get_intervention_org_units)
+        2. if matching_criteria is JSONLogic, intersect with org units whose MetricValues satisfy it
+           (match-all skips this step - every org unit qualifies)
         """
         if matching_criteria is None:
             return []
-        if isinstance(matching_criteria, dict) and matching_criteria.get("all"):
```
Bewi, line 188: I would still do it in two different logic.
Feels like you are fetching all org units, then fetching filtered org units.

I would keep the if is instance and return the full list.
And only filter if not is instance.
If you don't want to duplicate the flat and value_list you can use a variable.

models/scenario.py
```diff
@@ -180,30 +180,31 @@ def resolve_matched_org_units(account, matching_criteria):
...
         """Resolve the set of org unit ids this rule targets based on its matching mode."""
         if self.matching_criteria is None:
             return set(self.org_units_included)
-        is_match_all = isinstance(self.matching_criteria, dict) and self.matching_criteria.get("all")
-        matched = (
-            set(self.resolve_matched_org_units(self.scenario.account, self.matching_criteria))
-            if is_match_all
-            else set(self.org_units_matched)
-        )
+        matched = set(self.org_units_matched)
```
Bewi, line 206: Thanks for this <3

Bewi (changes requested):

Bewi (approved):
LGTM

PR #273, opened 2026-04-23
feature SNT-440: expose load_impact_org_unit_mappings in django admin

michiwend:

refactor the existing command into a reusable function and expose it in django admin with a file uploader form so we can avoid SSH-ing into prodicution environments

image
image


Bewi (approved):
LGTM, thanks for this :D

PR #274, opened 2026-04-23
SNT-433: Apply inflation on costs

Bewi:

Apply inflation when calculating costs

### Related JIRA tickets

SNT-433

## Changes

Instead of duplicating cost per year, apply the inflation using the formula: 

cost * (1+inflation_rate) ^ year

## How to test

Configure your costs for interventions in the interventions settings (You will need to select correct unit otherwise it will be ignored, you could just create the demo account to have all set up for you).
Create a scenario with a rule targetting that intervention. 
Checkout develop and run the budget.
Now checkout this branch and run again the budget. 
Make sure you si it increase by a bit based on the formula and the inflation of 0.03

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/api/budget/utils.py
+++ b/api/budget/utils.py
@@ -139,16 +139,22 @@ def build_cost_dataframe(account, start_year, end_year):
     # Convert to DataFrame
     df = pd.DataFrame(cost_lines_data)
 
-    # Duplicate data across scenario years
+    inflation_rate = float(budget_settings.inflation_rate)
+
+    # Duplicate data across scenario years, applying annual inflation compounding.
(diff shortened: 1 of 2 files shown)


michiwend (approved):
LGTM and fine to merge with one finding/thought

As far as I understand, the inflation calculation assumes that whatever number is stored in `unit_cost` reflects the cost as of a scenario's start year. So if `unit_cost` was entered for 2023 but a user runs a scenario starting in 2026, those 3 years of inflation between 2023 and 2026 are silently ignored.

Correct me if I'm wrong but otherwise worth a follow-up to record a reference year per cost line maybe?

PR #275, opened 2026-04-24
fix SNT-439: flaky input field height

michiwend:

This fixes a race condition of style sheets that makes our input fields render higher than initially after some navigation 

### Related JIRA tickets

SNT-439

## Changes

Initial state when loading the page
image

Styles break after some page navigation (inputs are higher)
image


Bewi (approved):
TIL :D

PR #276, opened 2026-04-24
Feature SNT-438: auto rule names and colour

michiwend:

- New rules get a random (unused) palette colour instead of black
- Rules without a name auto-generate names from interventions

### Related JIRA tickets

SNT-438

## Changes

Rules now accept empty names and the rule edit form in the front-end allows submitting rules without a name. When rules have no name, the front-end auto-generates a name based on a rule's intervention mix. This generated name is also shown to the user as the placeholder text in the rule name field.

This also made the rule name generation in the backend obsolete for imported scenarios. They will just be generated without names and the front-end shows the auto-names instead.

Initial colour in a new rule is now randomly picked based on non-used colours.

## How to test

- Create a new rule multiple times, a different initial colour should be set in the colour picker each time
- You can submit a rule with an empty name field
- The name field of a rule should show the combination of interventions as placeholder
- The list of rules without names should have their intervention mixes as names
- Rules with a custom name show the custom name, deleting the custom name makes them show the auto-name again

## Print screen / video

https://github.com/user-attachments/assets/3a7c4561-28d3-48b8-a34e-d553167f8c76


js/src/domains/planning/components/scenarioRule/ScenarioRulesPanel.tsx
```diff
@@ -20,18 +25,34 @@ export const ScenarioRulesPanel: FC<Props> = ({
...
 
     const { isEditing, toggleIsEditing } = usePlanningContext();
+    // Prefetches the palette so the colour is ready by the time the user
+    // clicks "create new rule".
+    const { data: palette } = useGetColors();
 
     const handleShowForm = useCallback(
         (rule?: ScenarioRule) => {
             setEditingRule(rule);
+            if (!rule) {
```
Bewi: I would put this in the form not here, purpose of this component should know as less as possible about form state

Bewi: Maybe in the useScenarioRuleFormState

Bewi (changes requested):
Overall it looks good.
I am just questionning the fact that we persist that. 

I can see the advantage, if it was generated and we modify the rule, it keeps updating it.
But I am also afraid of the confusion that if you come back to edit a rule and the name change without you notifying, you might wonder were is your rule when going back to the list.

Bewi (approved):
LGTM

PR #277, opened 2026-04-27
SNT-197 correctly handle the years in the budgeting population dataframe

Bewi:

Send target_population and make sure we import per year from OH

### Related JIRA tickets

SNT-197

## Changes

- Set year and is_population when importing from OH
- On intervention form, only display metric with is_population=true for target population 
- Send intervention target population columns config when calculating budget
- Fix a bug when importing metrics from OH and having custom metrics.

## How to test

You will need to setup an account for Cameroon. 
If you don't have the shapes already, you can go to [staging](https://explorer-staging.snt-toolbox.org) and download it. 
(Admin panel, Data source, select the used one and download geopackage). 
Then import it to your local newly created Cameroon account (make sure to tick the checkbox).
From there, you can go to the admin panel => account settings => Select `Region` for Focus org unit type and `District` for intervention org unit type.
You can also feed interventions for Cameroon: `docker compose run --rm iaso manage seed_interventions`

Now, you'll need your OpenHexa config: 
Go to the admin panel => OpenHexa Workspace: 
- select your account.
- slug: cmr-snt-process
- config: 
 
```
{
  "snt_results_dataset": "snt-result",
  "snt_population_dataset": "snt-dhis2-pop-transformation"
}
```

Now, you'll need your budget settings: 

Go to the admin panel, budget settings (under snt_malaria).
Mark local current as USD, 
Exchange are to 1 and inflation to 0.03

Time to import from open hexa by going to: `/snt_malaria/import_openhexa_metrics/`
Select your account and run it. 

It should work :D 

Go to interventions and add a cost for **CM** and select Other as unit type. 

- [ ] Go to `/snt_malaria/import_openhexa_metrics/`. Select your account and run it. It should work, verify the result log to see how many population metrics have been imported).
- [ ] Now going to **Data layer** you should have some metrics but NOT population metrics (Except for total population).
- [ ] You can add a custom metric and try to reimport from OH, it should still work.
- [ ] Go to **Interventions** and verify that you have costs for most of them.
- [ ] Go to Scenario, create a rule for CM and another one for ITN_Campaign, run the budget. You should have a cost for both of them (using the total population).
- [ ] Go to interventions and change the target population of CM, go back to the scenario and run the budget again, it should change.
- [ ] Verify that based on the year, the cost changes too as population grow and costs inflate.  

## Print screen / video

N/A

## Notes

You will need a PR from IASO for this to work: https://github.com/BLSQ/iaso/pull/2955

And One from SNT-Malaria budgeting: https://github.com/BLSQ/snt-malaria-budgeting/pull/49

## Doc

### DIFF
--- a/api/budget/utils.py
+++ b/api/budget/utils.py
@@ -16,7 +16,7 @@ from plugins.snt_malaria.models.budget_assumptions import BudgetAssumptions
 # TODO:
 # - think about the str() issue for org_unit_id
 # - think about error handling all the way to the frontend
-def build_population_dataframe(account, start_year, end_year):
+def build_population_dataframe(account, start_year, end_year, intervention_population_metric_codes=[]):
     """
     Build a population dataframe from MetricTypes with population data.
 
@@ -27,6 +27,8 @@ def build_population_dataframe(account, start_year, end_year):
       Populated from their respective MetricTypes
     """
     # Mapping from MetricType codes to DataFrame column names
+    # TODO: This should not be needed anymore, we should get this from metric types when we know how to define which one is a population metric type
+    # TODO: And the custom mapping should not exist anymore, if metric code doesn't match, they'll need to specify it manually from they frontend when creating the intervention
     metric_code_to_column = {
         "POPULATION": "pop_total",
         "POP_UNDER_5": "pop_0_5",
@@ -39,9 +41,15 @@ def build_population_dataframe(account, start_year, end_year):
     }
 
     pop_columns = list(metric_code_to_column.values())
+    pop_columns.extend(
+        intervention_population_metric_codes
+    )  # Add any additional population metric codes from interventions
 
     # Fetch all relevant MetricTypes for this account
     metric_types = MetricType.objects.filter(account=account, code__in=metric_code_to_column.keys())
+    metric_types = MetricType.objects.filter(
+        account=account, code__in=list(metric_code_to_column.keys()) + intervention_population_metric_codes
+    )
 
     if not metric_types.exists():
         raise ValidationError("No population MetricTypes found for this account")
@@ -64,6 +72,8 @@ def build_population_dataframe(account, start_year, end_year):
     # Convert to DataFrame
     df = pd.DataFrame(list(metric_values))
 
+    df.to_csv("population_metrics_raw.csv", index=False)  # Debug: Save raw metric values to CSV
+
     # Pivot the data so each metric type becomes a column
     df_pivoted = df.pivot_table(
         index=["org_unit_id"],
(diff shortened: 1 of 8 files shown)


api/budget/utils.py
```diff
@@ -64,6 +72,8 @@ def build_population_dataframe(account, start_year, end_year):
     # Convert to DataFrame
     df = pd.DataFrame(list(metric_values))
 
+    df.to_csv("population_metrics_raw.csv", index=False)  # Debug: Save raw metric values to CSV
```
michiwend: Is this intentional?

api/budget/utils.py
```diff
@@ -39,9 +41,15 @@ def build_population_dataframe(account, start_year, end_year):
...
     pop_columns = list(metric_code_to_column.values())
+    pop_columns.extend(
+        intervention_population_metric_codes
+    )  # Add any additional population metric codes from interventions
 
     # Fetch all relevant MetricTypes for this account
     metric_types = MetricType.objects.filter(account=account, code__in=metric_code_to_column.keys())
+    metric_types = MetricType.objects.filter(
+        account=account, code__in=list(metric_code_to_column.keys()) + intervention_population_metric_codes
+    )
```
michiwend, line 51: As far as I can tell, the first query (line 49) would be dead code

management/commands/support/metrics_importer.py
```diff
@@ -276,7 +282,11 @@ def _create_metric_values(self, csvfile, metric_types, org_units_mapping):
 
                         # Create the MetricValue
                         MetricValue.objects.create(
-                            metric_type=metric_type, org_unit=org_unit, value=value, string_value=string_value
+                            metric_type=metric_type,
+                            org_unit=org_unit,
+                            value=value,
+                            string_value=string_value,
+                            year=int(row.get("YEAR", 0)),
```
michiwend, line 289: maybe an edge case but default year to 0 seems a bit odd. Also the key expects all caps YEAR but the column filter above doesn't care about casing. Again, not sure if that can realistically happen.

api/budget/views.py
```diff
@@ -63,11 +63,20 @@ def create(self, request, *args, **kwargs):
             start_year,
             end_year,
         )
+
+        interventions = Intervention.objects.filter(intervention_category__account=request.user.iaso_profile.account)
```
michiwend, line 67: `interventions` is reassigned further down in line 81 which shadows this variable. Would suggest using a separate name or we can reuse the interventions?

michiwend (changes requested):

api/budget/utils.py
Bewi: Oupsy

management/commands/support/metrics_importer.py
Bewi: Yeah this bit is a bit weird, I kept what we had for verifying the columns but OH always provides them in upper case. And in the `row.get` we have to be case sensitive.

api/budget/views.py
Bewi: The second one is not needed, it should always be filtered by account.

michiwend (approved):

PR #278, opened 2026-04-28
FEAT: SNT-442 Add basic smoke tests on demo

Bewi:

Add basic smoke tests on demo

### Related JIRA tickets

SNT-442

## How to test

Checkout related iaso PR and modify your .env and .env.config (see readme).

Then run from iaso root: `npm run test:smoke`

2 should pass and 2 should be skipped (example tests)

## Print screen / video

N/A

## Notes

IASO PR: https://github.com/BLSQ/iaso/pull/2960

## Doc

N/A

### DIFF
--- a/js/__tests__/playwright/smoke/demoAccount.test.ts
+++ b/js/__tests__/playwright/smoke/demoAccount.test.ts
@@ -1,0 +1,27 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Demo account login smoke tests', () => {
+    test('login page appears', async ({ page }) => {
+        await page.goto('');
+
+        await expect(page).toHaveURL('/login/?next=%2Fdashboard%2Fhome%2F');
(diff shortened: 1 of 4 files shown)


js/__tests__/playwright/smoke/demoAccount.test.ts
```diff
@@ -0,0 +1,27 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Demo account login smoke tests', () => {
```
hugo093, line 4: good start , maybe in the future we could check that no error appears , like 
```javascript 
const expectNoErrors = async ({ page }: { page: Page }) => {
    await expect(
        page.locator('.error-container, .notistack-MuiContent-error'),
    ).toHaveCount(0);
    await expect(page.getByText('An exception occurred')).toHaveCount(0);
};
```

hugo093 (approved):
Just a minor detail but LGTM

PR #279, opened 2026-05-04
SNT-422 set coverage per year

Bewi:

Set coverage per year and per intervention assignment.

### Related JIRA tickets

SNT-422, SNT-423

## Changes

- Remove other cost assumptions from BudgetAssumption model.
- Add year to BudgetAssumption.
- Replace intervention code to intervention assignment on BudgetAssumption.
- Cleanup the Frontend form.
- Remove assignment creation as it is now only created via rules.
- Add a common_base test class and use it in most SNT tests

## How to test

Go to SNT, open a scenario, have some rules and run the budget (make sure you have some result there)
Go to the list of intervention assignments, open one, try to change the coverage for some years then:
 - close the side panel and re-open to make sure it was refreshed properly.
 - run the budget and verify that cost changed (total and per year)

Go to the admin and verify that you have as many budget assumptions as assignments.
Then change your rule and verify that assumptions count is still the same as assignments count.

## Print screen / video

## Notes

snt-malaria-budgeting [PR: SNT-422-set-coverage-per-year
](https://github.com/BLSQ/snt-malaria-budgeting/pull/51)

IASO: https://github.com/BLSQ/iaso/pull/2984

## Doc


models/budget_assumptions.py
```diff
@@ -5,31 +5,27 @@
...
-    buffer_mult = models.DecimalField(max_digits=3, decimal_places=2)
+    scenario = models.ForeignKey(Scenario, on_delete=models.CASCADE, related_name="budget_assumptions_set")
+    intervention_assignment = models.ForeignKey(
+        "snt_malaria.InterventionAssignment",
+        on_delete=models.CASCADE,
+        related_name="budget_assumptions",
+        null=True,
+        blank=True,
+    )
+    year = models.IntegerField(null=True, blank=True)
```
tdethier: I suggest:
- making both fields mandatory
- this will simplify the condition in the `UniqueConstraint` above
- we can get rid of previous `BudgetAssumptions` objects since they were no longer used: let's add a `RunPython` extra operation at the beginning of the migration, which is going to delete all objects

tdethier: Nitpick: you could pass `-500` to `year`, I would probably use a `PositiveSmallIntegerField`. And if you want to go further, you could define a validator that rejects values older than 2000? 2020? and greater than 2100?

models/scenario.py
```diff
@@ -259,6 +304,7 @@ def refresh_assignments(self, user: User, previous_assignments: dict[int, set[in
         InterventionAssignment.objects.bulk_create(intervention_assignments_to_create)
 
 
+# Deprecated replaces by BudgetAssumptions
```
tdethier, line 309: Maybe you might want to add a similar constraint on `coverage` in `BudgetAssumption`  (see below at `class Meta`)

api/budget_assumptions/serializers.py
```diff
@@ -19,70 +23,33 @@ def __init__(self, *args, **kwargs):
...
-            "touchpoints",
-            "tablet_factor",
-            "doses_per_child",
-        )
+        fields = ("id", "scenario", "intervention_assignment", "year", "coverage")
+
+
+class DefaultCostAssumptionsSerializer(serializers.Serializer):
+    def to_representation(self, instance):
+        print("Serializing default cost assumptions with instance:", instance)
```
tdethier:
```suggestion
```

js/src/domains/planning/contexts/PlanningContext.tsx
```diff
@@ -27,6 +29,7 @@ type PlanningContextType = {
     interventionCategories: InterventionCategory[];
     interventionAssignments: InterventionAssignmentResponse[];
     interventionPlans: InterventionPlan[];
+    defaultBudgetAssumptions: DefaultBudgetAssumptions | undefined;
```
tdethier: Shouldn't it be like this?
```suggestion
    defaultBudgetAssumptions?: DefaultBudgetAssumptions;
```

tdethier: And shouldn't it be a list?

js/src/domains/planning/hooks/useGetInterventionAssignments.tsx
```diff
@@ -19,7 +19,7 @@ const interventionAssignmentAPIRequest = (scenarioId: number) => {
 export const useGetInterventionAssignments = (
     scenarioId: number | undefined,
     select: (data: InterventionAssignmentResponse[]) => any = data => data,
-): UseQueryResult<InterventionPlan[], Error> => {
+): UseQueryResult<any, Error> => {
```
tdethier: I think you can keep the previous type

js/src/domains/planning/hooks/useGetInterventionAssignments.tsx
```diff
@@ -61,5 +61,8 @@ export const useGetInterventionPlans = (scenarioId: number | undefined) => {
 
             return acc;
         }, [] as InterventionPlan[]);
-    return useGetInterventionAssignments(scenarioId, select);
+    return useGetInterventionAssignments(scenarioId, select) as UseQueryResult<
```
tdethier: Do you need `as UseQueryResult`?

js/src/domains/planning/types/interventions.ts
```diff
@@ -50,24 +50,18 @@ export type InterventionOrgUnit = {
...
+    [interventionCode: string]: {
+        coverage: number;
+    };
+};
+
 export type BudgetAssumptions = {
     id: number;
-    intervention_code: string;
+    intervention_assignment: number;
+    year: number | null;
```
tdethier: Here as well, shouldn't it be this?


```suggestion
    year?: number;
```

api/budget/utils.py
```diff
@@ -247,24 +247,27 @@ def build_interventions_input(scenario):
...
     Build a map of budget assumptions for quick lookup by intervention code.
+    When year is provided, only assumptions for that year are applied.
+    Falls back to DEFAULT_COST_ASSUMPTIONS for any intervention code not explicitly overridden.
     """
-    assumptions = BudgetAssumptions.objects.filter(scenario=scenario)
+    assumptions = BudgetAssumptions.objects.filter(
+        scenario=scenario,
+        year__gte=scenario.start_year,
+        year__lte=scenario.end_year,
+        intervention_assignment__isnull=False,
```
tdethier: This filtering on assignment is no longer needed if you make the relation mandatory

api/budget_assumptions/serializers.py
```diff
@@ -95,3 +62,63 @@ def validate_scenario(self, value):
...
+        account = self.context["request"].user.iaso_profile.account
+        scenario = attrs.get("scenario")
+        assignment_ids = attrs.get("intervention_assignments") or []
+
+        # Deduplicate while keeping input order stable.
+        unique_assignment_ids = list(dict.fromkeys(assignment_ids))
+        assignments = list(
+            InterventionAssignment.objects.filter(
+                id__in=unique_assignment_ids,
+                scenario__account=account,
```
tdethier: I would filter on scenario instead of account, since you have it,

tdethier: This also means that you no longer need to fetch the account above

api/budget_assumptions/serializers.py
```diff
@@ -95,3 +62,63 @@ def validate_scenario(self, value):
...
+            )
+        )
+        assignments_by_id = {assignment.id: assignment for assignment in assignments}
+        missing_ids = [
+            assignment_id for assignment_id in unique_assignment_ids if assignment_id not in assignments_by_id
+        ]
+        if missing_ids:
```
tdethier, line 83: Nitpick: all of this works and is fine, but otherwise you could use sets instead of lists and use sets operators to get the difference

api/budget_assumptions/serializers.py
```diff
@@ -95,3 +62,63 @@ def validate_scenario(self, value):
...
+            raise serializers.ValidationError(
+                {"intervention_assignments": f"Invalid intervention assignment IDs: {missing_ids}"}
+            )
+
+        resolved_assignments = [assignments_by_id[assignment_id] for assignment_id in unique_assignment_ids]
+
+        if scenario and any(assignment.scenario_id != scenario.id for assignment in resolved_assignments):
+            raise serializers.ValidationError(
+                {"intervention_assignments": "All assignments must belong to the provided scenario."}
+            )
```
tdethier: You no longer need to do that if you directly filter on scenario

api/budget_assumptions/views.py
```diff
@@ -1,86 +1,61 @@
...
-    def _get_default_override(self, key, intervention_code):
-        prefix = f"{intervention_code}_anc" if intervention_code == "iptp" and key == "coverage" else intervention_code
-        return DEFAULT_COST_ASSUMPTIONS.get(f"{prefix}_{key}", 0)
+    @action(detail=False, methods=["post"])
+    def many(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        result = serializer.save()
+        response_serializer = BudgetAssumptionsReadSerializer(result, many=True)
+        return Response(response_serializer.data, status=status.HTTP_200_OK)
```
tdethier, line 58: I'm wondering if that's not why you still had duplicated queries when using FKs instead of a list of int.

Using another serializer for the return value is sometimes tricky, here's [how I did it for the create/update rules endpoint](https://github.com/BLSQ/snt-malaria/blob/d5f1721c6337e09425c0c8e18688ae057effce47/api/scenario_rules/views.py#L100)

Maybe we could test again with the `PrimaryKeyRelatedField` and tweak this block to check if we still have duplicated queries (if you have time)

api/budget_assumptions/views.py
```diff
@@ -1,86 +1,61 @@
...
-        assumptions_serializer = BudgetAssumptionsReadSerializer(all_assumptions, many=True)
+        serializer = BudgetAssumptionsReadSerializer(assumptions, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
 
-        return Response(assumptions_serializer.data, status=status.HTTP_200_OK)
+    @action(detail=False, methods=["get"])
+    def default(self, request, *args, **kwargs):
+        serializer = DefaultCostAssumptionsSerializer({}, many=False)
+        data = serializer.data
+        return Response(data, status=status.HTTP_200_OK)
```
tdethier: nitpick:

```suggestion
        return Response(serializer.data, status=status.HTTP_200_OK)
```

api/budget_assumptions/views.py
```diff
@@ -1,86 +1,61 @@
...
-
-        assumptions = self.get_queryset().filter(scenario_id=scenario.id)
-        interventions = (
-            Intervention.objects.prefetch_related("intervention_category__account")
-            .filter(intervention_category__account=self.request.user.iaso_profile.account)
-            .order_by("code")
-            .distinct("code")
-            .values("code")
+        query_serializer = BudgetAssumptionsQuerySerializer(
+            data=request.query_params, context=self.get_serializer_context()
```
tdethier: I would probably call `self.get_serializer()` instead, since you're manually instantiating the default serializer. Unless there's a specific reason that I missed

tests/api/budget_assumptions/test_serializers.py
```diff
@@ -1,510 +1,305 @@
...
-            buffer_mult="1.20",
-            doses_per_pw=2,
-            age_string="3-59",
-            pop_prop_3_11="0.10",
-            pop_prop_12_59="0.50",
-            monthly_rounds=3,
-            touchpoints=2,
-            tablet_factor="0.55",
-            doses_per_child=6,
+    def test_year_is_optional_and_nullable(self):
```
tdethier, line 110: I would maybe add a test where you provide a `year` value because there's none

tests/api/budget_assumptions/test_serializers.py
```diff
@@ -1,510 +1,305 @@
...
+        serializer = BudgetAssumptionsReadSerializer(assumption)
+        self.assertEqual(serializer.data["id"], assumption.id)
+        self.assertEqual(serializer.data["scenario"], self.scenario.id)
+        self.assertEqual(serializer.data["intervention_assignment"], self.assignment_a.id)
+        self.assertEqual(serializer.data["year"], 2025)
+        self.assertEqual(serializer.data["coverage"], "0.80")
+
+
+class BudgetAssumptionsUpdateSerializerTests(BudgetAssumptionsSerializerBaseTestCase):
+    def test_invalid_coverage_range(self):
```
tdethier, line 143: I would probably test with a negative value as well

tests/api/budget_assumptions/test_serializers.py
```diff
@@ -1,510 +1,305 @@
...
-        self.assertEqual(self.budget_assumption.doses_per_child, 62)
+        assumptions = BudgetAssumptions.objects.filter(scenario=self.scenario)
+        self.assertEqual(assumptions.count(), 3)  # Only the target row should be replaced, not the others
+
+        target = assumptions.get(year=2025, intervention_assignment=self.assignment_a)
+        same_scenario_other_year = assumptions.get(year=2026, intervention_assignment=self.assignment_a)
+        same_scenario_other_assignment = assumptions.get(year=2025, intervention_assignment=self.assignment_b)
+        other_scenario_row = BudgetAssumptions.objects.get(id=other_scenario_row.id)
+
+        self.assertEqual(str(target.coverage), "0.95")
```
tdethier: I would first try to fetch `target` based on its ID and expect a `BudgetAssumptions.DoesNotExist` exception to show that the object has been recreated.

And I would show that the ID is still the same for all other 3 objects, to show that they haven't been deleted

tests/api/budget_assumptions/test_serializers.py
```diff
@@ -1,510 +1,305 @@
...
+        self.assertEqual(str(same_scenario_other_assignment.coverage), "0.20")
+        self.assertEqual(str(same_scenario_other_year.coverage), "0.30")
+        self.assertEqual(str(other_scenario_row.coverage), "0.40")
+
+    def test_save_removing_assignment_deletes_related_assumptions(self):
```
tdethier: This test does not include a serializer, it should be part of model-level tests

tests/api/budget_assumptions/test_serializers.py
```diff
@@ -1,510 +1,305 @@
...
+        serializer = DefaultCostAssumptionsSerializer()
+        expected_defaults = DEFAULT_COST_ASSUMPTIONS
+        for key, value in serializer.data:
+            coverage_key = f"{key}_coverage"
+            self.assertIn(coverage_key, expected_defaults)
+            self.assertEqual(value["coverage"], expected_defaults[coverage_key])
 
-    def test_missing_scenario(self):
+
+class BudgetAssumptionsUpsertManySerializerTests(BudgetAssumptionsSerializerBaseTestCase):
```
tdethier, line 163: I would add tests for each exception that can be raised by the serializer's `validate()`

tests/common_base.py
```diff
@@ -0,0 +1,234 @@
...
+from iaso.models import Account, OrgUnit, OrgUnitType
+from iaso.test import APITestCase, TestCase
+from plugins.snt_malaria.models import Intervention, InterventionAssignment, InterventionCategory, Scenario
+
+
+class SNTMalariaTestMixin:
+    auto_create_account = True
+
+    def setUp(self):
+        super().setUp()
```
tdethier: This is a base class that doesn't have any inheritance, so it can't call `super().something` because it doesn't have any parent.

tdethier: So, if you create a test class that inherits from this mixin only, it will crash

tests/common_base.py
```diff
@@ -0,0 +1,234 @@
...
+        return {
+            "category_vaccination": category_vaccination,
+            "category_chemoprevention": category_chemoprevention,
+            "intervention_rts": intervention_rts,
+            "intervention_smc": intervention_smc,
+            "intervention_iptp": intervention_iptp,
+            "intervention_itn_campaign": intervention_itn_campaign,
+        }
+
+    def create_snt_default_interventions_setup(
```
tdethier, line 199: Nitpick: I wasn't sure I understood what this did, compared to the previous method. I would add more documentation and/or add something about assignments in the name

api/budget_assumptions/serializers.py
michiwend: Something to remove or replace with proper logging

tests/api/budget_assumptions/test_views.py
```diff
@@ -1,735 +1,411 @@
...
         response = self.client.get(f"{BASE_URL}?scenario={self.scenario.id}")
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
-    def test_get_budget_assumptions_invalid_scenario(self):
+    def test_get_default_budget_assumption(self):
         self.client.force_authenticate(user=self.user_with_full_perm)
-        response = self.client.get(f"{BASE_URL}?scenario=9999")  # Non-existent scenario ID
-        result = self.assertJSONResponse(response, status.HTTP_400_BAD_REQUEST)
-        self.assertIn("scenario", result)
+        response = self.client.get(f"{DEFAULT_URL}")
```
tdethier: Come on Claudy


```suggestion
        response = self.client.get(DEFAULT_URL)
```

tests/api/budget_assumptions/test_views.py
```diff
@@ -1,735 +1,411 @@
...
+            "intervention_assignments": [assignment.id for assignment in assignments],
+            "budget_assumptions": budget_assumptions,
+        }
+
     def test_get_budget_assumptions_unauthenticated(self):
         response = self.client.get(f"{BASE_URL}?scenario={self.scenario.id}")
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
-    def test_get_budget_assumptions_invalid_scenario(self):
+    def test_get_default_budget_assumption(self):
```
tdethier, line 126: I would add tests for perms/anon users as well for this specific endpoint

api/budget_assumptions/views.py
```diff
@@ -1,86 +1,61 @@
...
+            assumptions = assumptions.filter(year=year)
 
-        assumptions_serializer = BudgetAssumptionsReadSerializer(all_assumptions, many=True)
+        serializer = BudgetAssumptionsReadSerializer(assumptions, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
 
-        return Response(assumptions_serializer.data, status=status.HTTP_200_OK)
+    @action(detail=False, methods=["get"])
+    def default(self, request, *args, **kwargs):
+        serializer = DefaultCostAssumptionsSerializer({}, many=False)
```
tdethier: I don't think you need `many=False`

tests/api/budget_assumptions/test_views.py
```diff
@@ -1,735 +1,411 @@
...
-                        ("touchpoints", 0),
-                        ("tablet_factor", "0.00"),
-                        ("doses_per_child", 0),
-                    ]
-                ),
-            ],
+        self.assertEqual(len(result), 1)
+
+    def test_get_budget_assumptions_filters_by_year(self):
+        BudgetAssumptions.objects.create(
```
tdethier: I would store it in a variable, so that you can check it's ID in the returned payload

tests/api/budget_assumptions/test_views.py
```diff
@@ -1,735 +1,411 @@
...
-        unknown_assumption_id = 9999  # Assuming this ID does not exist in the test database
+        assumptions = BudgetAssumptions.objects.filter(
+            scenario=self.scenario,
+            intervention_assignment=self.assignment_a,
+            year=2025,
+        )
+        self.assertEqual(assumptions.count(), 1)
+        self.assertEqual(str(assumptions.first().coverage), "0.70")
+
+    def test_post_many_budget_assumptions_updates_only_given_scenario_assignments_and_years(self):
```
tdethier, line 376: You've already tested that at the serializer level, I think you could just check for existing IDs or not

models/scenario.py
```diff
@@ -62,13 +62,58 @@ def get_next_available_priority(self):
...
         new_assignments = {}
         for rule in self.rules.order_by("-priority"):
             rule.refresh_assignments(user, new_assignments)
+
+        self._restore_assumptions_for_all_assignments_(assumption_by_intervention)
         # Bump updated_at so the impact API cache (keyed on this timestamp) self-invalidates.
         self.save()
 
+    # As for the moment, assumptions are only managed per intervention, let's group them.
+    def _get_assumptions_by_intervention_and_year_(self):
```
tdethier: I would add unit tests for this

models/scenario.py
```diff
@@ -62,13 +62,58 @@ def get_next_available_priority(self):
...
+        )
+
+        assumptions_by_intervention = dict()
+        for intervention_id, year, coverage in assumptions:
+            if intervention_id not in assumptions_by_intervention:
+                assumptions_by_intervention[intervention_id] = {}
+            assumptions_by_intervention[intervention_id][year] = coverage
+        return assumptions_by_intervention
+
+    def _restore_assumptions_for_all_assignments_(self, assumptions_by_intervention):
```
tdethier: And for this as well

tests/api/intervention_categories/test_views.py
```diff
@@ -1,38 +1,35 @@
...
-        self.account, _, _, _ = self.create_account_datasource_version_project(
-            account_name="account", project_name="project", source_name="source"
-        )
-        self.user_with_perms, self.anon, self.user_no_perms = self.create_base_users(
-            self.account, [SNT_SCENARIO_FULL_WRITE_PERMISSION], "user"
-        )
+        super().setUp()
 
-        self.intervention_category_1 = InterventionCategory.objects.create(
+        self.account, self.user_with_perms = self.create_snt_account(name="Other Account")
```
tdethier: Nitpick: it's called `self.user_with_perms` but they don't have any perms

tests/api/scenarios/test_views.py
```diff
@@ -322,14 +301,6 @@ def test_scenario_create_same_name_different_account(self):
         self.assertEqual(new_scenario.account, other_account)
         self.assertEqual(new_scenario.created_by, other_user)
 
-    def test_scenario_create_unauthenticated(self):
```
tdethier, line 325: Why did you remove this test?

tdethier (changes requested):
General comments in any order:
- defining a common base class for SNT tests is a really good idea, thank you 👍 
- I think some tests are missing (see below)
- I think some of the new serializers could be simplified (see below)
- having a `BudgetAssumptions` per assignment per year might create a lot of values, I'm a bit scared about performance. I tried creating a rule for my Cameroon setup, and it crashed

Image

This weird SQL token error sounds like an error because of a massive `BudgetAssumptions` bulk create query, but I'm not sure. I tried to create a rule with few assignments, but it still caused the same issue

So, unfortunately, I couldn't test the feature 🫤

js/src/domains/planning/contexts/PlanningContext.tsx
Bewi: Not a list, this is an objet with intervention code and default coverage

models/budget_assumptions.py
Bewi: Did the PositiveSmallIntegerField but I'll skip the validation for now

js/src/domains/planning/types/interventions.ts
Bewi: It changes null to undefined but in this case it is fine

tests/api/scenarios/test_views.py
Bewi: Wasn't me :D

tdethier: 😢

michiwend (approved):

tdethier (approved):
LGTM, thanks for the changes 👍

PR #281, opened 2026-05-05
SNT-436 Cleanup nav permissions

Bewi:

Quick alignment with a fix merge in IASO, we don't need custom permission for always accessible menu items now.

### Related JIRA tickets

SNT-436

## Changes

There was a bug in IASO when given an empty permission list on a menu item would act as if nobody except admin has access.

As this has been fixed in IASO, this is a cleanup story to remove the role from SNT always accessible nav items.

## How to test

Have a non admin account and verify that navigation without permission are visible.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


tdethier (approved):
LGTM 👍

PR #284, opened 2026-05-07
[SNT-452] add throttling on public account creation

tdethier:

add throttling on public account creation

### Related JIRA tickets

SNT-452

## Changes

- added throttling on public account creation

## How to test
- set the `PUBLIC_ACCOUNT_THROTTLE_RATE` env var to the value of your choice (default value = `5/hour`; see [documentation here](https://www.django-rest-framework.org/api-guide/throttling/#setting-the-throttling-policy) to know which format to use)
- spam the `api/snt_malaria/account_setup/` endpoint, after crossing the limit you should get 429 errors

## Print screen / video

/

## Notes

- this requires [a IASO PR](https://github.com/BLSQ/iaso/pull/2998) to work
- even if the DRF documentation says this is not a security feature, it should work more or less as intended as it is coupled with a captcha (see other PR)

## Doc

/

### DIFF
--- a/api/account_setup/throttle.py
+++ b/api/account_setup/throttle.py
@@ -1,0 +1,5 @@
+from rest_framework.throttling import AnonRateThrottle
+
+
+class SNTAccountThrottle(AnonRateThrottle):
+    scope = "snt_public_account"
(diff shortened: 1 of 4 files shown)


Bewi (approved):
LGTM

PR #287, opened 2026-05-11
SNT-457 custom population layers

Bewi:

Add the possibilit to create population custom data layer,

### Related JIRA tickets

SNT-457

## Changes

- Moved metrics from planning to DataLayer domain.
- When creating a metric layer, add a toggle to specify if it is a population metric
- Add a year selector on data layer map if population
- Fetch data layers based on metric kind and don't show population metrics on planning

## How to test

Go to DataLayer and create a Layer for population by making sure to check that box under the variable.
Download the CSV template an fill up some values for your custom population layer (I used a base number and an increase ration of 0,02).
Upload it and select the year for which it should apply.
Modify your values and upload it for another year, as often as you need to (at least 3 years).

Now verify that when checking the layer on map you can select the year on top and that values match what you specified on the CSV.

Go to Interventions and set a cost for CM as we will use it later, make sure to select **Other** as unit type.

Once this is done, go to Scenario a create a Scenario with date ranges matching your population years.
When adding a rule verify that the population layer is **not**  visible from there matching criteria and in the map.
Add a rule for CM (easiest calculation: pop * coverage * cost). 
Verify that it is properly selected on your map. 

Then go to the budget and run it, it should gives you some result.
You can verify the cost if you sum up your population times cost.

## Print screen / video

N/A

## Notes

Related IASO PR: https://github.com/BLSQ/iaso/pull/3007

## Doc

N/A


js/src/domains/dataLayers/messages.ts
```diff
@@ -139,4 +139,17 @@ export const MESSAGES = defineMessages({
...
+    },
+    selectYear: {
+        id: 'iaso.snt_malaria.settings.dataLayers.selectYear',
+        defaultMessage: 'Select Year',
+    },
+    importCSVYearCaption: {
+        id: 'iaso.snt_malaria.settings.dataLayers.importCSVYearCaption',
+        defaultMessage:
+            'The year for which the metric values will be imported. Only apply to layer with population metrics.',
+    },
```
michiwend, line 154: not in the translations plus a small correction "applies"
```suggestion
    importCSVYearCaption: {
        id: 'iaso.snt_malaria.settings.dataLayers.importCSVYearCaption',
        defaultMessage:
            'The year for which the metric values will be imported. Only applies to layers with population metrics.',
    },
```

js/src/domains/dataLayers/dataLayerMap/DataLayerMapWrapper.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const valuesByYear: Record<number, MetricValue[]> = {};
+        const years = new Set<number>();
+        for (const metricValue of metricValues) {
+            const year = metricValue.year || 0;
+            if (!valuesByYear[year]) {
+                valuesByYear[year] = [];
+                years.add(year);
+            }
+            valuesByYear[year].push(metricValue);
+        }
```
michiwend: variable shadowing, see line 51
```suggestion
            const metricYear = metricValue.year || 0;
            if (!valuesByYear[metricYear]) {
                valuesByYear[metricYear] = [];
                years.add(metricYear);
            }
            valuesByYear[metricYear].push(metricValue);
        }
```

js/src/domains/dataLayers/MetricValuesImportDialog.tsx
```diff
@@ -1,16 +1,22 @@
...
     makeFullModal,
     useSafeIntl,
 } from 'bluesquare-components';
+import InputComponent from 'Iaso/components/forms/InputComponent';
 import { noOp } from 'Iaso/utils';
 import { useImportMetricValues } from './hooks/useImportMetricValues';
 import { MESSAGES } from './messages';
 
+const yearOptions = Array.from({ length: 20 }, (_, i) => {
+    const year = 2020 + i;
```
michiwend, line 16: hardcoded year, why not current year +/-?

michiwend (changes requested):
Works well, few findings down in the comments:

js/src/domains/dataLayers/MetricValuesImportDialog.tsx
Bewi: I don't know if we should allow past years or not, I'd rather have a fix number for now.
Mainly if you need to override your population of past year but bad luck, you cannot anymore.

michiwend: That's why I think we should have a window `currentYear - x` / `currentYear + x` but fine to keep it like this for now

michiwend (approved):

PR #291, opened 2026-05-20
SNT-476 budget create and migrate models

Bewi:

Update and create new models for budget

### Related JIRA tickets

SNT-476

## Changes

- Add a CostUnitType model and make changes to use it instead of a choice field
- Remove allowed_cost_unit_types from intervention
- Update intervention_seeder
- Add population layer and cost driver to cost lines.
- Create ScenarioYearlyCostAssignment model

What it is not about : 
- Adding api endpoints for new models
- Population layer is not set yet and ScenarioYearlyCostAssignment is not used yet.

## How to test
This should not change anything from user experience yet.

Regression tests:
- Verify that you can still created / modify costs and that unit types are listed properly.
- Verify that you can still modify budget assumptions on a scenario,
- Verify that budget calculation still runs properly and that result is the same as before.

## Print screen / video

N/A

## Notes

N/A

## Doc
N/A


api/intervention_cost_breakdown_line/serializers.py
```diff
@@ -27,35 +54,32 @@ class Meta:
...
-        required=True,
+    intervention = serializers.PrimaryKeyRelatedField(
+        queryset=Intervention.objects.all(),
     )
-
-
-class InterventionCostBreakdownLinesWriteSerializer(serializers.ModelSerializer):
-    intervention = serializers.PrimaryKeyRelatedField(queryset=Intervention.objects.all())
-    costs = CostsWriteSerializer(many=True)
+    unit_type = serializers.PrimaryKeyRelatedField(queryset=CostUnitType.objects.all(), required=False, allow_null=True)
```
michiwend: Before, `InterventionCostBreakdownLinesWriteSerializer.__init__` scoped intervention to the caller's account, and `test_create_breakdown_line_with_intervention_from_other_account` asserted `400`.

Your PR removes `POST` from `InterventionCostBreakdownLineViewSet`, so that test was downgraded to assert `405` which makes sense but now writing happens through `InterventionViewSet.update_details` (`PUT`) which uses `InterventionDetailWriteSerializer` --> `InterventionCostBreakdownLineWriteSerializer`. Both `intervention` and `unit_type` get written but neither serializer scopes their querysets to the caller's account.

I'm not really sure what the real-world risk of this is but I feel like this should not be possible. So we could have something like this:
```python
def _request_account(self):
    request = self.context.get("request")
    if request is None or not request.user.is_authenticated:
        return None
    return request.user.iaso_profile.account

def validate_intervention(self, value):
    account = self._request_account()
    if account is not None and value.intervention_category.account_id != account.id:
        raise serializers.ValidationError("Intervention does not belong to your account.")
    return value

def validate_unit_type(self, value):
    if value is None:
        return value
    account = self._request_account()
    if account is not None and value.account_id != account.id:
        raise serializers.ValidationError("Cost unit type does not belong to your account.")
    return value
```

models/cost_breakdown.py
```diff
@@ -39,8 +43,22 @@ class Meta:
         choices=InterventionCostBreakdownLineCategory.choices,
         default=InterventionCostBreakdownLineCategory.OTHER,
     )
-    unit_type = models.CharField(
-        max_length=50, choices=InterventionCostUnitType.choices, default=InterventionCostUnitType.OTHER
+    unit_type = models.ForeignKey(
+        "snt_malaria.CostUnitType",
+        on_delete=models.CASCADE,
+        related_name="cost_breakdown_lines",
+    )
```
michiwend, line 50: `unit_type` is defined as `NOT NULL` here in the model.

But  `InterventionCostBreakdownLineWriteSerializer` has
```python
unit_type = serializers.PrimaryKeyRelatedField(..., required=False, allow_null=True)
```
So it accepts `NULL` which would fail and also the read serialiser defences against `NULL` which only makes sense if it can be `NULL` :
```python
def get_unit_type_label(self, obj):
    return obj.unit_type.name if obj.unit_type_id else None
```

models/cost_breakdown.py
```diff
@@ -39,8 +43,22 @@ class Meta:
         choices=InterventionCostBreakdownLineCategory.choices,
         default=InterventionCostBreakdownLineCategory.OTHER,
     )
-    unit_type = models.CharField(
-        max_length=50, choices=InterventionCostUnitType.choices, default=InterventionCostUnitType.OTHER
+    unit_type = models.ForeignKey(
+        "snt_malaria.CostUnitType",
+        on_delete=models.CASCADE,
```
michiwend: Are we sure we want a `CASCADE` here? Its a bit dangerous IMO and a safer approach here could be `PROTECT`, so you would not be able to delete a `unit_type` if it is still in use by other cost lines. Also see `population_layer` you have `SET_NULL` there.

api/intervention_cost_breakdown_line/serializers.py
```diff
@@ -1,6 +1,33 @@
...
+        for item in validated_data:
+            lines.append(
+                InterventionCostBreakdownLine(
+                    created_by=request_user,
+                    updated_by=request_user,
+                    **item,
+                )
+            )
+
+        return InterventionCostBreakdownLine.objects.bulk_create(lines)
```
michiwend: Are we sure about deleting + re-creating all cost lines on every `PUT`? Two things make me a bit nervous here:
- Even a single rename regenerates every sibling's id
- `ScenarioYearlyCostAssignment.cost_line` is `CASCADE`, so this approach wipes the cost assignments silently

Also: migration 0042 populates `ScenarioYearlyCostAssignment` from `BudgetAssumptions`... so the first `PUT` to `update_details` after that migration silently wipes the data we just migrated

models/cost_breakdown.py
```diff
@@ -29,6 +29,10 @@ class InterventionCostBreakdownLineCategory(models.TextChoices):
         SUPPORTIVE = "Supportive", _("Supportive")
         OTHER = "Other", _("Other")
 
+    class CostDriver(models.TextChoices):
+        POPULATION = "population", _("Population")
+        FIX_COST = "fix_cost", _("Fixed cost")
```
michiwend: knit pick: fix vs fixed

tests/api/intervention_cost_breakdown_lines/test_serializers.py
michiwend, line 97: All those tests pass for the wrong reason, the new serializer has no `costs` field

tests/api/intervention_cost_breakdown_lines/test_views.py
```diff
@@ -127,26 +118,17 @@ def test_create_breakdown_line_with_intervention_from_other_account(self):
 
         self.client.force_authenticate(user=self.user_write)
         response = self.client.post(self.BASE_URL, data, format="json")
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
         self.assertEqual(InterventionCostBreakdownLine.objects.count(), 3)  # from setup
```
michiwend: still has the old payload shape but its anyway not needed for a `HTTP_405_METHOD_NOT_ALLOWED` test. We can just remove the payload

This is true for all `test_create_*`

michiwend (changes requested):
Have a few findings down in the files 👇

api/intervention_cost_breakdown_line/serializers.py
Bewi: Good catch, I'll scope the query set for both of them

models/cost_breakdown.py
Bewi: Don't know, it kinda make sense to me that if you remove a unit_type all associated costs are removed.
It shouldn't be that dangerous and it prevents manging this case.
However, I can set it as PROTECT so it will be block, but we'll need to make sure user feedback is properly managed.

As for population_layer, I would say the same, we can PROTECT it but then we need to make sure we provide proper feedback to the user.

api/intervention_cost_breakdown_line/serializers.py
Bewi: Damn, this wasn't an issue before but here side effects are to big. 
I'll implement a better diff here.

api/intervention_cost_breakdown_line/serializers.py
```diff
@@ -9,19 +10,38 @@ def update(self, queryset, validated_data):
...
-                    updated_by=request_user,
-                    **item,
+            line_id = item.get("id")
+            if line_id and line_id in existing_lines:
+                line = existing_lines[line_id]
+                for attr, value in item.items():
+                    setattr(line, attr, value)
+                lines_to_update.append(line)
+                lines_to_delete.discard(line_id)
+            else:
```
michiwend, line 25: I suggest adding something like
```python
item.pop("id", None)
```
or maybe change the condition to check for empty id, otherwise you can essentially create new ids

tests/api/intervention_cost_breakdown_lines/test_serializers.py
```diff
@@ -17,49 +18,223 @@ def setUp(self):
...
-            "costs": [{"name": "test", "unit_cost": 15, "category": invalid_category_id}],
+            "name": "test",
+            "unit_cost": 15,
+            "category": "NotACategory",
+            "unit_type": self.unit_type_other.id,
         }
-        serializer = InterventionCostBreakdownLineSerializer(data=data, context=self.context)
+        serializer = InterventionCostBreakdownLineWriteSerializer(data=data, context=self.context)
         self.assertFalse(serializer.is_valid())
         self.assertIn("category", serializer.errors)
```
michiwend, line 97: I suggest we just collapse them into one because now they all do the same:

```python
def test_write_methods_are_not_allowed(self):
    self.client.force_authenticate(user=self.user_write)
    for method in ("post", "put", "patch", "delete"):
        response = getattr(self.client, method)(self.BASE_URL, {}, format="json")
        self.assertEqual(
            response.status_code,
            status.HTTP_405_METHOD_NOT_ALLOWED,
            f"{method.upper()} should be rejected, got {response.status_code}",
        )
    self.assertEqual(InterventionCostBreakdownLine.objects.count(), 3)
```

tests/api/intervention_cost_breakdown_lines/test_serializers.py
```diff
@@ -15,68 +15,226 @@ def setUp(self):
...
+        )
+        self.assertTrue(serializer.is_valid(), serializer.errors)
+        serializer.save()
+
+        with self.assertRaises(InterventionCostBreakdownLine.DoesNotExist):
+            InterventionCostBreakdownLine.objects.get(id=self.cost_line2.id)
+
+        with self.assertRaises(ScenarioYearlyCostAssignment.DoesNotExist):
+            ScenarioYearlyCostAssignment.objects.get(id=yearly_assignment.id)
+
```
michiwend, line 250: Some additional tests for the cross-account fix:

```suggestion

def test_intervention_from_other_account_is_rejected(self):
    data = {
        "intervention": self.other_intervention.id,
        "name": "x",
        "unit_cost": "1.00",
        "category": "Procurement",
        "unit_type": self.unit_type_other.id,
    }
    serializer = InterventionCostBreakdownLineWriteSerializer(data=data, context=self.context)
    self.assertFalse(serializer.is_valid())
    self.assertIn("intervention", serializer.errors)

def test_unit_type_from_other_account_is_rejected(self):
    data = {
        "intervention": self.intervention_chemo_iptp.id,
        "name": "x",
        "unit_cost": "1.00",
        "category": "Procurement",
        "unit_type": self.other_unit_type.id,
    }
    serializer = InterventionCostBreakdownLineWriteSerializer(data=data, context=self.context)
    self.assertFalse(serializer.is_valid())
    self.assertIn("unit_type", serializer.errors)
 
```

api/intervention_cost_breakdown_line/serializers.py
```diff
@@ -1,6 +1,53 @@
...
+                        created_by=request_user,
+                        updated_by=request_user,
+                        **item,
+                    )
+                )
+        with transaction.atomic():
+            if lines_to_update:
+                InterventionCostBreakdownLine.objects.bulk_update(
+                    lines_to_update,
+                    fields=["name", "unit_cost", "unit_type", "category", "intervention", "updated_by"],
```
michiwend, line 38: This needs to be hand-maintained, otherwise we drop fields silently. The next PR already adds new fields. One way to keep it in sync could be:
```suggestion
                    fields = [
                        f.name for f in InterventionCostBreakdownLine._meta.concrete_fields
                        if f.name not in ("id", "created_at", "created_by")
                    ]
```

michiwend (approved):
Nothing blocking, just a few additional things. Feel free to fix them:

api/intervention_cost_breakdown_line/serializers.py
Bewi: Yeah, I'll don't do that now, I don't think we already do this anywhere and I'd rather stick with the same way of working for this

tests/api/intervention_cost_breakdown_lines/test_serializers.py
Bewi: Feels like this one is on the wrong file, I'll apply that on the view tests

PR #293, opened 2026-05-21
SNT-479 add new fields in cost line form

Bewi:

Add form input to set Cost line population_layer 

### Related JIRA tickets

SNT-479

## Changes

- Add population layer in put to cost line form.
- Remove target_population from intervention form (not yet from the model).
- Moved some front end model to proper domain.
- Migrate target_population to population_layer.
- Add a message to explain a bit how the population layer works.

## How to test

Before getting this branch, go to interventions and assign some population.
Checkout the branch, run the migration and go to interventions.
Each cost line of that intervention should have the same population as before.
Try to modify it, it should only apply to the cost line you changed it for.
Check that the cost_driver is set to either "Population" if you selected a layer or "fixed_cost" if you selected nothing.

## Print screen / video

N/A

## Notes

This breaks doesn't break the budget calculation BUT the population per cost line is not reflected and it will use population set on the intervention.

## Doc


api/intervention_cost_breakdown_line/serializers.py
michiwend: `population_layer` is missing in the fields for bulk update now. Also `cost_driver` but as far as I understood, `cost_driver` gets derived in `validate` and is not a user input, which makes me think whether it should be a model field in the first place or rather a computed property?

michiwend, line 27: carried over from last PR... `updated_by` is not a serializer field, so it isn't in item. If we want to update the user we should do something like
```python
line.updated_by = request_user
```

js/src/domains/interventions/types/interventionCostBreakdownLine.ts
```diff
@@ -0,0 +1,11 @@
+export type InterventionCostBreakdownLine = {
+    name: string;
+    category: string;
+    category_label: string;
+    unit_type: string;
+    unit_type_label: string;
+    unit_cost: string; // using string to avoid float precision issues
```
michiwend: that is an FK now, right? So we can make it a number

js/src/domains/interventions/types/interventionCostBreakdownLine.ts
```diff
@@ -0,0 +1,11 @@
...
+    id: number;
+    intervention: number;
+    population_layer: string | null;
```
michiwend: also here, should be `number | null`

michiwend (changes requested):

api/intervention_cost_breakdown_line/serializers.py
Bewi: Yeah I've been wondering myself for the cost_driver.
Current usage feels like a computed property, I did it like this 'cause I can see comming that it will be used for more than this, but this is just a feeling.

Bewi: Nice catch

michiwend (approved):

PR #294, opened 2026-05-22
Feature: SNT-464, SNT-411, SNT-409 new delta map implementation and refactoring

michiwend:

Improvements for the Impact Delta Map. Covers three tickets, see below

### Related JIRA tickets

SNT-464 - Support different metrics in the impact comparison map
SNT-411 - The difference map shows difference where there are none
SNT-409 - The difference scale should be based on absolute values rather than relative

## Changes

* Central definition and configuration of metrics and how they are supposed to be formatted
* Show all available metrics in dropdown
* Add cost as a metric
* Linear, continuous scale instead of buckets
* Option to share the scale between all or only the current two scenarios
* Require at least two scenarios before rendering the map

Central definition of all known metric fields. All types now are derived from this single source of truth:
```ts
export const ImpactMetricKey = {
    DirectDeaths: 'direct_deaths',
    Cases: 'number_cases',
    SevereCases: 'number_severe_cases',
    PrevalenceRate: 'prevalence_rate',
} as const;

export type ImpactMetricKey =
    (typeof ImpactMetricKey)[keyof typeof ImpactMetricKey];
```

Central configuration for metrics which determines number format, label and scale direction for each value in one place. This is currently only local to the differences map but can later be lifted so the entire compare module uses the same.
```ts
const METRIC_CONFIG: Record<
    MetricKey,
    {
        label: IntlMessage;
        format: FormatNumberOptions;
        positiveIsGreen: boolean;
    }
> = {
    [MetricKey.DirectDeaths]: {
        label: MESSAGES.deaths,
        format: { notation: 'compact', maximumFractionDigits: 1 },
        positiveIsGreen: false,
    },
    [MetricKey.Cases]: {
        label: MESSAGES.impactCases,
        format: { notation: 'compact', maximumFractionDigits: 1 },
        positiveIsGreen: false,
    },
    [MetricKey.SevereCases]: {
        label: MESSAGES.impactSevereCases,
        format: { notation: 'compact', maximumFractionDigits: 1 },
        positiveIsGreen: false,
    },
    [MetricKey.PrevalenceRate]: {
        label: MESSAGES.impactDifferenceMetricPrevalenceRate,
        format: { style: 'percent', maximumFractionDigits: 2 },
        positiveIsGreen: false,
    },
    [MetricKey.OrgUnitTotalCost]: {
        label: MESSAGES.impactDifferenceMetricOrgUnitCost,
        format: { notation: 'compact', maximumFractionDigits: 1 },
        positiveIsGreen: false,
    },
};
```

## Print screen / video

image

Shared scale switches between a scale across all scenarios vs only the current two. A shared scale helps with comparability between three scenarios as it stays stable when switching the second scenario:

image

Show only available metrics:

image
image

New linear scale
image


js/src/domains/compareCustomize/components/maps/ImpactDifferencesMap.tsx
```diff
@@ -67,77 +114,96 @@ export const ImpactDifferencesMap: FC = () => {
...
-        const values = Array.from(deltaMap.values());
-        return buildDivergingScale(Math.min(...values), Math.max(...values));
-    }, [deltaMap]);
+        const allDeltas: number[] = [];
+        if (useSharedScale) {
+            for (const m of deltaMapsByScenarioId.values()) {
+                for (const v of m.values()) allDeltas.push(v);
+            }
+        } else {
+            for (const v of deltaMap.values()) allDeltas.push(v);
```
Bewi: would go for allDeltas = Array.from(deltaMap.values())
One less loop.

js/src/domains/compareCustomize/components/maps/ImpactDifferencesMap.tsx
```diff
@@ -67,77 +114,96 @@ export const ImpactDifferencesMap: FC = () => {
...
+        [deltaMap, scale, intl, metricConfig],
     );
 
     const baselineScenario = scenarios.find(s => s.id === baselineScenarioId);
     const selectedComparison = comparisonScenarios.find(
         s => s.id === selectedComparisonId,
     );
     const hasData = deltaMap.size > 0;
-    const dataKey = `${selectedComparisonId ?? 'none'}_${deltaMap.size}`;
+    const dataKey = [
```
Bewi: Maybe a use memo to recompile this only when one of the element change ?

Bewi (approved):
WaE and LGTM. 
Only few small comments.

PR #295, opened 2026-05-27
Scenario Yearly Cost Assignment API

Bewi:

API for Scenario yearly cost assignments

### Related JIRA tickets

SNT-488

## Changes

Add views, serializers and tests for Scenario yearly cost assignments

## How to test

Nothing much to test atm as nothing connects to it yet.

## Doc

### DIFF
--- a/api/scenario_yearly_cost_assignment/permissions.py
+++ b/api/scenario_yearly_cost_assignment/permissions.py
@@ -1,0 +1,11 @@
+from plugins.snt_malaria.api.scenarios.views import ScenarioPermission
+from plugins.snt_malaria.models import ScenarioYearlyCostAssignment
+
+
+class ScenarioYearlyCostAssignmentPermission(ScenarioPermission):
+    def has_object_permission(self, request, view, obj: ScenarioYearlyCostAssignment):
+        # # Override this one as scenario permission check this only for deletion
+        if obj.scenario.is_locked:
+            return False
+
+        return super().has_object_permission(request, view, obj.scenario, extra_safe_methods=[])
--- a/api/scenario_yearly_cost_assignment/serializers.py
+++ b/api/scenario_yearly_cost_assignment/serializers.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from rest_framework import serializers
 
 from plugins.snt_malaria.models import InterventionCostBreakdownLine, Scenario, ScenarioYearlyCostAssignment
@@ -64,11 +65,10 @@ class ScenarioYearlyCostAssignmentUpsertSerializer(serializers.Serializer):
             ).filter(intervention__intervention_category__account=account)
         return fields
 
-    def validate(self, attrs):
-        intervention = attrs["intervention"]
-        if not intervention.cost_breakdown_lines.exists():
-            raise serializers.ValidationError(f"Intervention {intervention.id} does not have any cost breakdown line.")
-        return attrs
+    def validate_intervention(self, value):
+        if not value.cost_breakdown_lines.exists():
+            raise serializers.ValidationError(f"Intervention {value.id} does not have any cost breakdown line.")
+        return value
 
     def validate_scenario(self, value):
         if value.is_locked:
@@ -77,6 +77,7 @@ class ScenarioYearlyCostAssignmentUpsertSerializer(serializers.Serializer):
 
     # Save info, if cost_line is provided, update or create the assignment for this cost line
     # if not provided update or create the assignment for all population cost lines of the intervention
+    @transaction.atomic
     def save(self, **kwargs):
         intervention = self.validated_data["intervention"]
         year = self.validated_data["year"]
--- a/api/scenario_yearly_cost_assignment/serializers.py
+++ b/api/scenario_yearly_cost_assignment/serializers.py
@@ -1,0 +1,103 @@
+from rest_framework import serializers
+
+from plugins.snt_malaria.models import InterventionCostBreakdownLine, Scenario, ScenarioYearlyCostAssignment
+from plugins.snt_malaria.models.intervention import Intervention
+
+
+class ScenarioYearlyCostAssignmentSerializer(serializers.ModelSerializer):
+    cost_line = serializers.PrimaryKeyRelatedField(queryset=InterventionCostBreakdownLine.objects.none())
+    scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none())
+
+    class Meta:
+        model = ScenarioYearlyCostAssignment
+        fields = ["id", "scenario", "year", "cost_line", "value"]
+        read_only_fields = ["id"]
+
+    def get_fields(self):
+        fields = super().get_fields()
+        request = self.context.get("request")
+        if request:
+            account = request.user.iaso_profile.account
+            fields["scenario"].queryset = Scenario.objects.filter(account=account)
+            fields["cost_line"].queryset = InterventionCostBreakdownLine.objects.select_related(
+                "intervention__intervention_category"
+            ).filter(intervention__intervention_category__account=account)
+        return fields
+
+
+class ScenarioYearlyCostAssignmentQuerySerializer(serializers.Serializer):
+    scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none())
+
+    def get_fields(self):
+        fields = super().get_fields()
+
+        request = self.context.get("request")
+        if request:
+            account = request.user.iaso_profile.account
+            fields["scenario"].queryset = Scenario.objects.filter(account=account)
+        return fields
+
+
+class ScenarioYearlyCostAssignmentUpsertSerializer(serializers.Serializer):
+    intervention = serializers.PrimaryKeyRelatedField(
+        queryset=InterventionCostBreakdownLine.objects.none(), required=True
+    )
+    scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none(), required=True)
+    cost_line = serializers.PrimaryKeyRelatedField(
+        queryset=InterventionCostBreakdownLine.objects.none(), required=False
+    )
+    year = serializers.IntegerField(max_value=2100, min_value=2000, required=True)
+    value = serializers.DecimalField(max_digits=19, decimal_places=2, required=True)
+
+    def get_fields(self):
+        fields = super().get_fields()
+        request = self.context.get("request")
+        if request:
+            account = request.user.iaso_profile.account
+
+            fields["intervention"].queryset = Intervention.objects.select_related("intervention_category").filter(
+                intervention_category__account=account
+            )
+            fields["scenario"].queryset = Scenario.objects.select_related("account").filter(account=account)
+            fields["cost_line"].queryset = InterventionCostBreakdownLine.objects.select_related(
+                "intervention__intervention_category"
+            ).filter(intervention__intervention_category__account=account)
+        return fields
+
+    def validate(self, attrs):
+        intervention = attrs["intervention"]
+        if not intervention.cost_breakdown_lines.exists():
+            raise serializers.ValidationError(f"Intervention {intervention.id} does not have any cost breakdown line.")
+        return attrs
+
+    def validate_scenario(self, value):
+        if value.is_locked:
+            raise serializers.ValidationError("Cannot assign yearly cost to a locked scenario.")
+        return value
+
+    # Save info, if cost_line is provided, update or create the assignment for this cost line
+    # if not provided update or create the assignment for all population cost lines of the intervention
+    def save(self, **kwargs):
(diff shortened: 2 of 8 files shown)


api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+from plugins.snt_malaria.models import InterventionCostBreakdownLine, Scenario, ScenarioYearlyCostAssignment
+from plugins.snt_malaria.models.intervention import Intervention
+
+
+class ScenarioYearlyCostAssignmentSerializer(serializers.ModelSerializer):
+    cost_line = serializers.PrimaryKeyRelatedField(queryset=InterventionCostBreakdownLine.objects.none())
+    scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none())
+
+    class Meta:
+        model = ScenarioYearlyCostAssignment
```
tdethier, line 13: Nitpick: I still don't like the name of that class 😅

api/scenario_yearly_cost_assignment/views.py
```diff
@@ -0,0 +1,44 @@
...
+from plugins.snt_malaria.api.scenario_yearly_cost_assignment.serializers import (
+    ScenarioYearlyCostAssignmentQuerySerializer,
+    ScenarioYearlyCostAssignmentSerializer,
+    ScenarioYearlyCostAssignmentUpsertSerializer,
+)
+from plugins.snt_malaria.models import ScenarioYearlyCostAssignment
+
+
+class ScenarioYearlyCostAssignmentViewSet(viewsets.ModelViewSet):
+    serializer_class = ScenarioYearlyCostAssignmentSerializer
```
tdethier: You don't need to set it if you write `get_serializer_class`
```suggestion
```

api/scenario_yearly_cost_assignment/views.py
```diff
@@ -0,0 +1,44 @@
...
+from plugins.snt_malaria.api.scenario_yearly_cost_assignment.permissions import ScenarioYearlyCostAssignmentPermission
+from plugins.snt_malaria.api.scenario_yearly_cost_assignment.serializers import (
+    ScenarioYearlyCostAssignmentQuerySerializer,
+    ScenarioYearlyCostAssignmentSerializer,
+    ScenarioYearlyCostAssignmentUpsertSerializer,
+)
+from plugins.snt_malaria.models import ScenarioYearlyCostAssignment
+
+
+class ScenarioYearlyCostAssignmentViewSet(viewsets.ModelViewSet):
```
tdethier, line 13: Nitpick: I like setting `http_method_names` explicitly here in order to know which operations are allowed. It doesn't change much, but for instance here both PUT and PATCH are allowed

api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+        fields = ["id", "scenario", "year", "cost_line", "value"]
+        read_only_fields = ["id"]
+
+    def get_fields(self):
```
tdethier, line 17: Why did you do that in `get_fields` instead of `__init__`? Not saying that it's bad, I'm just wondering 😅

api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+            fields["cost_line"].queryset = InterventionCostBreakdownLine.objects.select_related(
+                "intervention__intervention_category"
+            ).filter(intervention__intervention_category__account=account)
+        return fields
+
+
+class ScenarioYearlyCostAssignmentQuerySerializer(serializers.Serializer):
+    scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none())
+
+    def get_fields(self):
```
tdethier, line 32: Same comment for all serializers

api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+            fields["cost_line"].queryset = InterventionCostBreakdownLine.objects.select_related(
+                "intervention__intervention_category"
+            ).filter(intervention__intervention_category__account=account)
+        return fields
+
+    def validate(self, attrs):
+        intervention = attrs["intervention"]
+        if not intervention.cost_breakdown_lines.exists():
+            raise serializers.ValidationError(f"Intervention {intervention.id} does not have any cost breakdown line.")
+        return attrs
```
tdethier: Since you're validating a single field, I would put this in `validate_intervention` instead

api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def validate_scenario(self, value):
+        if value.is_locked:
+            raise serializers.ValidationError("Cannot assign yearly cost to a locked scenario.")
+        return value
+
+    # Save info, if cost_line is provided, update or create the assignment for this cost line
+    # if not provided update or create the assignment for all population cost lines of the intervention
+    def save(self, **kwargs):
```
tdethier, line 91: I would probably make this atomic, I can imagine issues if there's an error in the middle of the `for` loop

api/scenario_yearly_cost_assignment/permissions.py
```diff
@@ -0,0 +1,11 @@
+from plugins.snt_malaria.api.scenarios.views import ScenarioPermission
```
tdethier, line 1: You need to add an `__init__.py` file in this directory

tests/api/scenario_yearly_cost_assignments/test_serializers.py
```diff
@@ -0,0 +1,331 @@
...
+        self.population_line_2 = InterventionCostBreakdownLine.objects.create(
+            intervention=self.intervention,
+            unit_type=self.unit_type,
+            name="Population line 2",
+            category="Operational",
+            cost_driver=InterventionCostBreakdownLine.CostDriver.POPULATION,
+            unit_cost=15,
+            created_by=self.user,
+        )
+        self.fix_cost_line = InterventionCostBreakdownLine.objects.create(
```
tdethier: fixed? 😅

tests/api/scenario_yearly_cost_assignments/test_serializers.py
```diff
@@ -0,0 +1,331 @@
...
+
+        current_scenario_assignments = ScenarioYearlyCostAssignment.objects.filter(
+            scenario=self.scenario,
+            year=2026,
+        )
+        self.assertEqual(current_scenario_assignments.count(), 3)
+
+        updated_line_1 = current_scenario_assignments.get(cost_line=self.population_line_1)
+        updated_line_2 = current_scenario_assignments.get(cost_line=self.population_line_2)
+        unchanged_fix_line = current_scenario_assignments.get(cost_line=self.fix_cost_line)
```
tdethier: fixed? 😅

tests/api/scenario_yearly_cost_assignments/test_serializers.py
```diff
@@ -0,0 +1,331 @@
...
+                "intervention": empty_intervention.id,
+                "scenario": self.scenario.id,
+                "year": 2026,
+                "value": "99.00",
+            },
+            context=self._context_for(self.user),
+        )
+
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("non_field_errors", serializer.errors)
```
tdethier: This will change if you place the validation in `validate_intervention`

tests/api/scenario_yearly_cost_assignments/test_serializers.py
```diff
@@ -0,0 +1,331 @@
...
+                "cost_line": self.other_account_cost_line.id,
+                "value": "12.00",
+            },
+            context=self._context_for(self.user),
+        )
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("cost_line", serializer.errors)
+
+
+class ScenarioYearlyCostAssignmentUpsertManySerializerTests(ScenarioYearlyCostAssignmentSerializerBaseTestCase):
```
tdethier, line 137: I would add tests for:
- checking limits of what can (not) be passed to `value`
- passing a `cost_line` ID that belongs to another account
- passing a `cost_line` ID that does not belong to the `intervention` ID that was passed

tests/api/scenario_yearly_cost_assignments/test_views.py
```diff
@@ -0,0 +1,254 @@
...
+            intervention=self.other_account_intervention,
+            unit_type=self.other_account_unit_type,
+            name="Other line",
+            category="Procurement",
+            cost_driver=InterventionCostBreakdownLine.CostDriver.POPULATION,
+            unit_cost=15,
+            created_by=self.other_account_user,
+        )
+
+    def _create_assignment(self, scenario, cost_line, year=2026, value="5.00"):
```
tdethier: This is part of the reason why I really hate this name: when I see this function I expect it to create an `InterventionAssignment` object

tests/api/scenario_yearly_cost_assignments/test_views.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def test_create_forbids_no_access_same_account(self):
+        response = self._post_create_payload(self.user_no_perm, self.scenario, self.population_line_1)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    # This shouldn't work on locked scenario
+    def test_create_allows_full_access_locked_scenario(self):
+        response = self._post_create_payload(self.user_with_full_perm, self.locked_scenario, self.population_line_1)
+        result = self.assertJSONResponse(response, status.HTTP_201_CREATED)
```
tdethier: The comment is a TODO for another PR? Or you forgot to implement this?

tests/api/scenario_yearly_cost_assignments/test_views.py
```diff
@@ -0,0 +1,254 @@
...
+from plugins.snt_malaria.models import InterventionCostBreakdownLine, ScenarioYearlyCostAssignment
+from plugins.snt_malaria.models.cost_unit_type import CostUnitType
+from plugins.snt_malaria.permissions import SNT_SCENARIO_BASIC_WRITE_PERMISSION, SNT_SCENARIO_FULL_WRITE_PERMISSION
+from plugins.snt_malaria.tests.common_base import SNTMalariaAPITestCase
+
+
+BASE_URL = "/api/snt_malaria/scenario_yearly_cost_assignments/"
+
+
+class ScenarioYearlyCostAssignmentAPITestCase(SNTMalariaAPITestCase):
```
tdethier, line 12: I would add tests for:
- checking permissions for deleting lines
- deleting lines that belong to the user account
- deleting lines that do not belong to the user account

tdethier (changes requested):
Overall it looks good, I left a few comments and nitpicks 👍

I'm mostly requesting changes because of the `__init__.py` file and to add a few missing tests

tests/api/scenario_yearly_cost_assignments/test_views.py
Bewi: Function name is off for this one, I'll rename it, no reason for such hate :D

api/scenario_yearly_cost_assignment/serializers.py
Bewi: I had a problem when __init__ was reached to soon for nested serializers and the request was never defined.
Having it in the fields fixes that problem :)

tests/api/scenario_yearly_cost_assignments/test_views.py
Bewi: I've added only one as deletion is not allowed from API for this one.

tests/api/scenario_yearly_cost_assignments/test_views.py
Bewi: This is actually covered by the seriliazer, as I am reusing the scenario permissions.
But it feels wrong so I might override permissions

Bewi: Actually, to prevent duplicate fetch of scenario, I'll keep it in serializer, I've updated the comment.

api/scenario_yearly_cost_assignment/serializers.py
tdethier: Oh I see. That's interesting. This shows that's probably how we should do it, instead of `__init__`.

I guess we don't need to change that everywhere, but we should probably do it like this from now on

tests/api/scenario_yearly_cost_assignments/test_views.py
tdethier: I was talking about the model name, not the method, don't worry :sweat_smile:

tdethier (approved):
LGTM 👍

PR #297, opened 2026-05-28
SNT-481 Add internal budget calculation

Bewi:

Create internal budget service to calculate costs

### Related JIRA tickets

SNT-481

## Changes

Add a new budget service + tests
Not used yet anywhere else.

## How to test

Not really testable atm, only verify that tests are running properly.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/services/budget.py
+++ b/services/budget.py
@@ -1,0 +1,371 @@
+from collections import defaultdict
+from dataclasses import dataclass, field
+from decimal import Decimal
+
+from iaso.models import MetricValue
+from plugins.snt_malaria.models import BudgetSettings, InterventionCostBreakdownLine, ScenarioYearlyCostAssignment
+
+
+@dataclass
+class BudgetBreakdownItem:
+    category: str
+    cost_class: str
+    total_cost: Decimal = Decimal("0.0")
+    quantity: Decimal = Decimal("0.0")
+
+
+@dataclass
+class BudgetInterventionItem:
+    id: int
+    code: str
+    type: str
+    total_cost: Decimal = Decimal("0.0")
+    total_pop: Decimal = Decimal("0.0")
+    quantity: Decimal = Decimal("0.0")
+    cost_breakdown: list[BudgetBreakdownItem] = field(default_factory=list)
+
+
+@dataclass
+class BudgetOrgUnitInterventionItem:
+    id: int
+    code: str
+    type: str
+    total_cost: Decimal = Decimal("0.0")
+    quantity: Decimal = Decimal("0.0")
+    cost_breakdown: list[BudgetBreakdownItem] = field(default_factory=list)
+
+
+@dataclass
+class BudgetOrgUnitItem:
+    org_unit_id: int
+    total_cost: Decimal = Decimal("0.0")
+    quantity: Decimal = Decimal("0.0")
+    interventions: list[BudgetOrgUnitInterventionItem] = field(default_factory=list)
+
+
+@dataclass
+class BudgetYearResult:
+    year: int
+    total_cost: Decimal
+    quantity: Decimal
+    interventions: list[BudgetInterventionItem]
+    org_units_costs: list[BudgetOrgUnitItem]
+    category_costs: list[BudgetBreakdownItem]
+
+
+@dataclass
+class BudgetLineRow:
+    org_unit_id: int
+    intervention_id: int
+    category: str
+    cost_class: str
+    population: Decimal
+    quantity: Decimal
+    total_cost: Decimal
+
+
+class BudgetCalculationService:
+    """Compute scenario budget using internal population-driven formula.
+
+    Implemented formula:
+    population * yearly_assignment_value * cost_unit_ratio * cost_line_unit_cost
+
+    Only population-driven cost lines are processed in this first version.
+    """
+
+    buffer = Decimal("1.1")
+
+    def __init__(self, scenario):
+        self.scenario = scenario
+        self.start_year = scenario.start_year
+        self.end_year = scenario.end_year
+
+        self.assignments = list(
+            scenario.intervention_assignments.select_related("intervention", "org_unit")
+            .all()
+            .order_by("org_unit_id", "intervention_id")
+        )
+        self.intervention_meta_by_id = {}
+        for assignment in self.assignments:
+            self.intervention_meta_by_id[assignment.intervention_id] = {
+                "code": assignment.intervention.code,
+                "type": assignment.intervention.short_name,
+            }
+
+        intervention_ids = {a.intervention_id for a in self.assignments}
+        self.population_cost_lines = list(
+            InterventionCostBreakdownLine.objects.filter(
+                intervention_id__in=intervention_ids,
+                cost_driver=InterventionCostBreakdownLine.CostDriver.POPULATION,
+            )
+            .select_related("population_layer", "unit_type", "intervention")
+            .order_by("intervention_id", "id")
+        )
+
+        self.cost_lines_by_intervention_id = defaultdict(list)
+        for line in self.population_cost_lines:
+            self.cost_lines_by_intervention_id[line.intervention_id].append(line)
+
+        org_unit_ids = {a.org_unit_id for a in self.assignments}
+        metric_type_ids = {
+            line.population_layer_id for line in self.population_cost_lines if line.population_layer_id is not None
+        }
+
+        metric_values = MetricValue.objects.filter(
+            org_unit_id__in=org_unit_ids,
+            year__gte=self.start_year,
+            year__lte=self.end_year,
+            metric_type_id__in=metric_type_ids,
+        ).values("org_unit_id", "year", "metric_type_id", "value")
+        self.population_by_key = {
+            (row["org_unit_id"], row["year"], row["metric_type_id"]): Decimal(str(row["value"]))
+            for row in metric_values
+        }
+
+        yearly_assignments = ScenarioYearlyCostAssignment.objects.filter(
+            scenario=scenario,
+            year__gte=self.start_year,
+            year__lte=self.end_year,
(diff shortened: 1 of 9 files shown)


services/budget.py
```diff
@@ -0,0 +1,371 @@
+from collections import defaultdict
+from dataclasses import dataclass, field
+from decimal import Decimal
+
+from iaso.models import MetricValue
+from plugins.snt_malaria.models import BudgetSettings, InterventionCostBreakdownLine, ScenarioYearlyCostAssignment
+
+
+@dataclass
+class BudgetBreakdownItem:
```
tdethier: I would maybe put all dataclasses in another file in services, like `utils.py` or `dataclasses.py`

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+        self.intervention_meta_by_id = {}
+        for assignment in self.assignments:
+            self.intervention_meta_by_id[assignment.intervention_id] = {
+                "code": assignment.intervention.code,
+                "type": assignment.intervention.short_name,
+            }
+
+        intervention_ids = {a.intervention_id for a in self.assignments}
+        self.population_cost_lines = list(
```
tdethier: You don't use it later, you don't need to store it in `self`

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+            .order_by("org_unit_id", "intervention_id")
+        )
+        self.intervention_meta_by_id = {}
+        for assignment in self.assignments:
+            self.intervention_meta_by_id[assignment.intervention_id] = {
+                "code": assignment.intervention.code,
+                "type": assignment.intervention.short_name,
+            }
+
+        intervention_ids = {a.intervention_id for a in self.assignments}
```
tdethier: You already go through assignments on line 89, you populate `intervention_ids` at the same time

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            .select_related("population_layer", "unit_type", "intervention")
+            .order_by("intervention_id", "id")
+        )
+
+        self.cost_lines_by_intervention_id = defaultdict(list)
+        for line in self.population_cost_lines:
+            self.cost_lines_by_intervention_id[line.intervention_id].append(line)
+
+        org_unit_ids = {a.org_unit_id for a in self.assignments}
```
tdethier: Here as well, you could populate `org_unit_ids` directly in your loop on line 89

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+        metric_type_ids = {
```
tdethier: You already go through `population_cost_lines` on line 106, you could populate `metric_type_ids` there as well

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+        self.population_by_key = {
+            (row["org_unit_id"], row["year"], row["metric_type_id"]): Decimal(str(row["value"]))
+            for row in metric_values
+        }
+
+        yearly_assignments = ScenarioYearlyCostAssignment.objects.filter(
+            scenario=scenario,
+            year__gte=self.start_year,
+            year__lte=self.end_year,
+            cost_line_id__in=[line.id for line in self.population_cost_lines],
```
tdethier: You could store the cost line IDs in another variable while you go through cost lines earlier. Or you don't turn `self.population_cost_lines` into a list and call the ORM to return a list of IDs (with something like `.values_list("id", flat=True)`

services/budget/budget_calculation.py
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        budget_settings = BudgetSettings.objects.filter(account=scenario.account).first()
+        self.inflation_rate = Decimal(str(budget_settings.inflation_rate)) if budget_settings else Decimal("0")
+
+    def calculate_all_years(self):
+        return [self.calculate_year(year) for year in range(self.start_year, self.end_year + 1)]
+
+    def calculate_year(self, year):
+        """Calculate the budget for a given year, based on the population-driven formula and the scenario data.
```
tdethier, line 96: Kudos for the documentation everywhere, it would be impossible to follow otherwise 👍

services/budget/budget_calculation.py
```diff
@@ -0,0 +1,371 @@
...
+        for assignment in self.assignments:
+            intervention = assignment.intervention
+            org_unit_id = assignment.org_unit_id
+
+            for line in self.cost_lines_by_intervention_id.get(intervention.id, []):
+                if line.population_layer_id is None:
+                    continue
+
+                population = self.population_by_key.get((org_unit_id, year, line.population_layer_id), Decimal("0"))
+                if population <= 0:
```
tdethier, line 190: Can you compare a `Decimal("0")` with `0`?

services/budget.py
```diff
@@ -0,0 +1,371 @@
...
+            org_units_costs=org_units_costs,
+            category_costs=category_costs,
+        )
+
+    def _compute_breakdown_line_rows(self, year):
+        """
+        Compute the raw cost breakdown lines for a given year, without any aggregation, to be used as input for the budget calculation.
+        Calculation is based on the population-driven formula, only processing cost lines with population as cost driver
+        And skipping lines with missing population or yearly assignment values.
+        Formula quantity: population * yearly_assignment_value * cost_unit_ratio
```
tdethier: Sorry to be bothering you again with this, but here we use 2 different types of assignments in the same method and it's just very confusing 😅

tests/services/test_budget.py
```diff
@@ -0,0 +1,257 @@
...
+        result = service.calculate_year(2025)
+
+        # quantity = pop * yearly_value * ratio
+        # district_1: 1000 * 1.2 * 0.5 = 600
+        # district_2: 2000 * 1.2 * 0.5 = 1200
+        # total quantity = 1800
+        self.assertEqual(result.quantity, 1800.0)
+
+        # cost = quantity * unit_cost(2.0) * buffer(1.1)
+        # district_1 = 1200 * 1.1 = 1320, district_2 = 2400 * 1.1 = 2640, total = 3960
```
tdethier, line 142: Thanks for writing down these details 👍

tests/services/test_budget.py
```diff
@@ -0,0 +1,257 @@
...
+        self.assertEqual(intervention.quantity, 1800.0)
+        self.assertEqual(intervention.total_pop, 3000.0)
+        self.assertEqual(intervention.total_cost, 3960.0)
+        self.assertEqual(len(intervention.cost_breakdown), 1)
+        self.assertEqual(intervention.cost_breakdown[0].category, "Procurement")
+        self.assertEqual(intervention.cost_breakdown[0].total_cost, 3960.0)
+
+        self.assertEqual(len(result.org_units_costs), 2)
+        district_1_result = next(o for o in result.org_units_costs if o.org_unit_id == self.district_1.id)
+        district_2_result = next(o for o in result.org_units_costs if o.org_unit_id == self.district_2.id)
```
tdethier: Nitpick: since you order values based on ID, you should be able to simply fetch `result.org_units_costs[0]` for the first org unit that's created

tests/services/test_budget_comparison.py
```diff
@@ -0,0 +1,515 @@
...
+from plugins.snt_malaria.tests.common_base import SNTMalariaTestCase
+
+
+class BudgetCalculationComparisonTestCase(SNTMalariaTestCase):
+    def setUp(self):
+        super().setUp()
+
+        budgeting_module = importlib.import_module("snt_malaria_budgeting")
+        self.BudgetCalculator = budgeting_module.BudgetCalculator
+        self.default_cost_assumptions = budgeting_module.DEFAULT_COST_ASSUMPTIONS
```
tdethier: Why don't you directly import them?

tests/services/test_budget.py
```diff
@@ -0,0 +1,257 @@
...
+    InterventionAssignment,
+    InterventionCostBreakdownLine,
+    ScenarioYearlyCostAssignment,
+)
+from plugins.snt_malaria.models.cost_unit_type import CostUnitType
+from plugins.snt_malaria.services.budget import BudgetCalculationService
+from plugins.snt_malaria.tests.common_base import SNTMalariaTestCase
+
+
+class BudgetCalculationServiceTestCase(SNTMalariaTestCase):
```
tdethier, line 15: Ideally, I would have split the `calculate_budget` tests from the utility methods tests so that it's easier to test and check for coverage for each of them. I think it's mostly fine as it is, but it's a bit difficult to determine if you went through all cases of each utility method like this 😅

tests/services/test_budget_comparison.py
```diff
@@ -0,0 +1,515 @@
...
+                float(legacy_places_with_cost[0]["total_cost"]),
+                places=3,
+            )
+            self.assertAlmostEqual(
+                float(internal_result.total_cost),
+                float(sum(item["total_cost"] for item in legacy_interventions)),
+                places=3,
+            )
+
+    def test_compare_iptp_quantification(self):
```
tdethier, line 215: Nice way to compare both 👍

tdethier (approved):
This looks good, nice job documenting everything 👍 

I made some suggestions to avoid going through multiple loops of the same data

services/budget.py
Bewi: I moved both budget.py and dataclasses to a subfolder to scope it better.

services/budget.py
Bewi: I tried to give a more descriptive name in this documentation :D

services/budget/budget_calculation.py
Bewi: Yep, decimal support comparison with float and int

services/budget.py
Bewi: Indeed, I also moved the metric_type_ids to the same loop.

PR #300, opened 2026-06-01
SNT-478 modify budget view

Bewi:

Implement new budget view and use internal budget calculation function.

### Related JIRA tickets

SNT-478

## Changes

- Add Budget table
- Remove budget view
- Use internal calculation 
- Use ScenarioYearlyCostASsignments instead of Budget assumptions   

## How to test

Go to SNT Malaria, have a scenario with some rules / assignments.

Go to the budget tab, you should see list of interventions with a cost (or not if you didn't run the budget yet).

Cost should be per year. 

Un-collapsing an intervention line should show a list of cost line with inputs for the coverage.
Coverage is saved on change (after 500ms debounce).
Changing the coverage doesn't refresh the budget yet so don't forget to press the run budget button.

Remove an intervention from the rule and add it back, it keeps previous coverage.

Hovering the progress bar on the right should show a tooltip with % of budget it represent.

## Print screen / video

image

## Notes

We will need to clean up BudgetAssumptions but this will be done in another PR to keep this one more focus.

## Doc

N/A


services/budget/budget_calculation.py
```diff
@@ -137,11 +140,13 @@ def calculate_year(self, year):
...
             org_unit_intervention_totals[row.org_unit_id][intervention_id]["quantity"] += row.quantity
+            org_unit_intervention_breakdowns[row.org_unit_id][intervention_id][row.category]["id"] = row.cost_line_id
             org_unit_intervention_breakdowns[row.org_unit_id][intervention_id][row.category]["total_cost"] += (
                 row.total_cost
             )
             org_unit_intervention_breakdowns[row.org_unit_id][intervention_id][row.category]["quantity"] += row.quantity
 
+            category_totals[row.category]["id"] = row.cost_line_id
             category_totals[row.category]["total_cost"] += row.total_cost
             category_totals[row.category]["quantity"] += row.quantity
```
michiwend, line 163: we group the breakdown by category here but only keep a single `cost_line_id` (last one wins). So, when you have multiple cost lines within the same category, it just dumps the total into one cost line and puts all other to 0. Same pattern at lines 134 and 143. Confirmed with this test:

Image

Image

Image

services/budget/dataclasses.py
```diff
@@ -1,46 +1,50 @@
...
+
+class BudgetBaseModel(BaseModel):
+    # This base model is used to ensure that all Decimal fields are serialized as floats in JSON responses
+    @field_serializer("*", when_used="json", check_fields=False)
+    def serialize_decimal(self, value):
+        return float(value) if isinstance(value, Decimal) else value
+
+
+class BudgetBreakdownItem(BudgetBaseModel):
+    id: int
```
michiwend: tiny robustness thing, `id` is required but the defaultdicts start with `id=None` in `budget_calculations.py`

js/src/domains/planning/hooks/useSaveScenarioYearlyCostAssignment.tsx
```diff
@@ -0,0 +1,46 @@
+import { patchRequest, postRequest } from 'bluesquare-components';
+import { UseMutationResult } from 'react-query';
+import { useSnackMutation } from 'Iaso/libs/apiHooks';
+
+type ScenarioYearlyCostAssignmentPayload = {
+    id?: number;
+    scenario: number;
+    intervention: number;
```
michiwend: `PlanningContext.saveYearlyCoverage` calls this without intervention, so the POST body at line 38 ships `intervention: undefined` but here its required

api/budget/views.py
```diff
@@ -77,60 +76,22 @@ def create(self, request, *args, **kwargs):
...
+            year_result = budget_service.calculate_year(year)
+            budgets.append(year_result)
 
         budget = Budget.objects.create(
             scenario=scenario,
             name=f"Budget for {scenario.name}",
-            cost_input=cost_df.astype(str).to_dict(orient="records"),
-            population_input=population_df.astype(str).to_dict(orient="records"),
+            cost_input=cost_df.astype(str).to_dict(orient="records"),  # TODO Remove this
+            population_input=population_df.astype(str).to_dict(orient="records"),  # TODO Remove this
```
michiwend, line 92: should be removed

js/src/components/LinearProgress.tsx
```diff
@@ -0,0 +1,66 @@
...
+            {values.map((value, index) => (
+                <Tooltip
+                    key={colors[index] + value}
+                    title={getTooltipTitle(index)}
+                    arrow
+                >
+                    <Box
+                        sx={{
+                            ...style.segment,
+                            width: `${(value / max) * 100}%`,
```
michiwend: division by 0 with no budget yet

js/src/domains/planning/components/budgeting/BudgetTable.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    return acc;
+                },
+                {} as Record<number, (BudgetIntervention & { year: number })[]>,
+            ),
+        [budgets],
+    );
+
+    useEffect(() => {
+        const rows = [] as BudgetRowData[];
+        const totalCosts = {
```
michiwend: shadows the state variable

michiwend (changes requested):
Really nice work ❤️ 

One finding that is not part of the change set but `getCostBreakdownChartData` in the `cost-utils.tsx` is dead code and also references `costLine.cost` which was renamed.

few other things down in the comms

api/budget/views.py
Bewi: Yes, this will be part of the cleanup task. 
I am still not sure if we should remove this or populate it.
Anyway, in my next PR it is replaced by an empty object

services/budget/budget_calculation.py
Bewi: Nice catch, we don't need to group this per category now.
Leftover from my first implementation.

services/budget/dataclasses.py
Bewi: I'll mark it optional as there is no way to initialize it.
I wonder why it didn't fail though

michiwend (approved):

PR #303, opened 2026-06-04
SNT-482 auto refresh budget

Bewi:

Auto refresh budget when related data changes.

### Related JIRA tickets

SNT-482

## Changes

- Remove old budget usage
- Moved budget refresh to a service. 
- Call budget refresh on API when cost line, rule or scenario yearly cost assignment change

## How to test

Go to SNT with a budget and try to :
- Change coverage
- Change rule (preview not covered)
- Change cost line in intervention settings

Make sure that budget is refreshed for each of them.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


api/interventions/views.py
```diff
@@ -33,4 +35,18 @@ def update_details(self, request, pk):
...
+            .filter(intervention=intervention)
+            .values_list("scenario", flat=False)
+            .distinct()
+        )
+
+        scenarios = intervention.intervention_category.account.scenario_set.filter(id__in=scenarioIds)
+        for scenario in scenarios:
+            budget_service = BudgetCalculationService(scenario)
+            budget_service.calculate_and_save_all_years(self.request.user)
+
```
michiwend, line 52: I think we should do this transactional. If one of the calculate fails, we end up having inconsistent budget. We could just make the whole handler `@transaction.atomic`

api/scenario_rules/views.py
```diff
@@ -116,6 +124,8 @@ def perform_destroy(self, instance):
         scenario = instance.scenario
         super().perform_destroy(instance)
         scenario.refresh_assignments(self.request.user)
+        budget_service = BudgetCalculationService(scenario)
+        budget_service.calculate_and_save_all_years(self.request.user)
```
michiwend, line 129: same concern about `@transaction.atomic`

services/budget/budget_calculation.py
```diff
@@ -89,6 +94,19 @@ def __init__(self, scenario):
...
+            scenario=self.scenario,
+            name=f"Budget for {self.scenario.name}",
+            cost_input={},
+            population_input={},
+            assumptions={},
+            results=[budget_result.model_dump(mode="json") for budget_result in all_years_results],
+            created_by=user,
+            updated_by=user,
+        )
+
```
michiwend, line 109: why do we want to keep old budgets? Just wondering about the amount of stale budget data this creates over time given that this runs on every tiny edit now

api/scenario_yearly_cost_assignment/views.py
```diff
@@ -36,6 +37,24 @@ def list(self, request, *args, **kwargs):
...
+        scenario = serializer.validated_data["scenario"]
+        budget_service = BudgetCalculationService(scenario)
+        budget_service.calculate_and_save_all_years(self.request.user)
+
+        return response
+
+    def perform_update(self, serializer):
+        response = super().perform_update(serializer)
+        # After creating/updating the ScenarioYearlyCostAssignment, we need to recalculate the budget for the related scenario to reflect the changes in the assigned costs
+        scenario = serializer.validated_data["scenario"]
```
michiwend: On partial_update, `scenario` may not be in `validated_data`. maybe `serializer.instance.scenario` instead?

api/interventions/views.py
```diff
@@ -33,4 +35,18 @@ def update_details(self, request, pk):
...
+        # Refresh budget for all scenarios with at least 1 assignment of that intervention
+        # As it might impact cost lines.
+        scenarioIds = (
+            InterventionAssignment.objects.prefetch_related("scenario")
+            .filter(intervention=intervention)
+            .values_list("scenario", flat=False)
+            .distinct()
+        )
+
+        scenarios = intervention.intervention_category.account.scenario_set.filter(id__in=scenarioIds)
```
michiwend: `prefetch_related("scenario")` is a no-op on a `values_list` with `flat=False` also we can change the casing if we're already touching it:

```suggestion
        scenario_ids = InterventionAssignment.objects
        .filter(intervention=intervention)
        .values_list("scenario_id", flat=True)
        .distinct().
        
        scenarios = intervention.intervention_category.account.scenario_set.filter(id__in=scenario_ids)
```

michiwend (changes requested):
Really nice work! I just found a few non-critical things

services/budget/budget_calculation.py
Bewi: Yeah, we can discuss this, it was before we thought they would like to compare between two states.
Not sure we'll need it anymore but let's move this in another story then.

michiwend (approved):

PR #304, opened 2026-06-05
SNT-485 handle fix costs

Bewi:

Handle cost lines with fixed cost.

### Related JIRA tickets

SNT-485

## Changes

- Change input to accept non percentage value
- Transform value to percentage on API instead of FE
- Select content of input on focus
- Cleaned up logic for coverage as we are updating per cost line, we don't need to batch update.
- 

## How to test

Go to SNT interventions.
Add a cost with no population.
Go to a scenario where there is at least one assignment for that intervention.
On budget view, you should see that cost line under the intervention cost. 
By default it should be 0.
Changing the quantity will update the cost, you can verify the value (cost * buffer (1,1) * cost_unit_ratio * coverage).

## Print screen / video

## Notes

## Doc

### DIFF
--- a/api/scenario_yearly_cost_assignment/serializers.py
+++ b/api/scenario_yearly_cost_assignment/serializers.py
@@ -1,8 +1,8 @@
-from django.db import transaction
+import math
+
 from rest_framework import serializers
 
 from plugins.snt_malaria.models import InterventionCostBreakdownLine, Scenario, ScenarioYearlyCostAssignment
-from plugins.snt_malaria.models.intervention import Intervention
 
 
 class ScenarioYearlyCostAssignmentSerializer(serializers.ModelSerializer):
@@ -25,6 +25,16 @@ class ScenarioYearlyCostAssignmentSerializer(serializers.ModelSerializer):
             ).filter(intervention__intervention_category__account=account)
         return fields
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        value = float(instance.value)
+        if instance.cost_line.cost_driver == InterventionCostBreakdownLine.CostDriver.POPULATION:
+            value *= 100
+
+        data["value"] = f"{math.floor(value)}"
+
+        return data
+
 
 class ScenarioYearlyCostAssignmentQuerySerializer(serializers.Serializer):
     scenario = serializers.PrimaryKeyRelatedField(queryset=Scenario.objects.none())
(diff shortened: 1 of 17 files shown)


api/scenario_yearly_cost_assignment/views.py
```diff
@@ -59,5 +53,15 @@ def post(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         result = serializer.save()
-        response_serializer = ScenarioYearlyCostAssignmentSerializer(result, many=True)
+        self._recalculate_budget(serializer.validated_data["scenario"])
+        response_serializer = ScenarioYearlyCostAssignmentSerializer(result, context=self.get_serializer_context())
+        return Response(response_serializer.data, status=status.HTTP_200_OK)
```
michiwend: `post` is not a standard handler on a `ModelViewSet` so this code never runs but instead the default `create` is called and that means we don't update the budget. The previous `perform_create` was the right thing imo. This is also confirmed by the create tests which assert a `201` but this `post` would return `200`

api/scenario_yearly_cost_assignment/serializers.py
```diff
@@ -25,6 +25,16 @@ def get_fields(self):
             ).filter(intervention__intervention_category__account=account)
         return fields
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        value = float(instance.value)
+        if instance.cost_line.cost_driver == InterventionCostBreakdownLine.CostDriver.POPULATION:
+            value *= 100
+
+        data["value"] = f"{math.floor(value)}"
```
michiwend: float(Decimal("0.29")) * 100 = 28.999… --> floor --> 28. Tests only cover 20%/150% (which happen to survive)

api/scenario_yearly_cost_assignment/views.py
```diff
@@ -59,5 +53,15 @@ def post(self, request, *args, **kwargs):
...
+
+    def partial_update(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        result = serializer.save()
+        self._recalculate_budget(serializer.validated_data.get("scenario", instance.scenario))
+        response_serializer = ScenarioYearlyCostAssignmentSerializer(result, context=self.get_serializer_context())
         return Response(response_serializer.data, status=status.HTTP_200_OK)
```
michiwend: This basically turns `perform_update` into dead code... the default update flow (`partial_update` > `update` > `perform_update`) never runs because this override does the save + recalc inline

together with my comment above, I think we can drop both the `post` and `partial_update` overrides, keep only `create`/`update` overrides for the response shape (default `partial_update` delegates to `update`, so `PATCH` is covered), and put the side effects in `perform_create`/`perform_update`

michiwend (changes requested):

api/scenario_yearly_cost_assignment/serializers.py
Bewi: I'll add a test for this, I'll round instead of floor.

michiwend (approved):
LGTM

PR #305, opened 2026-06-05
Feature SNT-480: manage cost unit types

michiwend:

Manage cost unit types from new settings module

### Related JIRA tickets

SNT-480
SNT-486

## Changes

The changes also include model changes which rename `ratio` to `value` and introduce an `invert_value` flag from which `ratio` is now computed using a model property. This way `ratio` doesn't change for the budget function but the user can enter the value more flexibly and doesn't need to do the math themselves.

The frontend uses the inversion flag to display the value either as  "x per" or "Covers x". So if for example a bale covers 90 people, it can be reflected directly and not through 0.11. This also prevents loss of precision.

## Print screen / video

image


api/cost_unit_types/views.py
```diff
@@ -0,0 +1,48 @@
...
+            instance.delete()
+        except ProtectedError:
+            raise MethodNotAllowed(
+                self.request.method,
+                _("Cannot delete this cost unit because it is used by one or more cost lines."),
+            )
+
+    @staticmethod
+    def _raise_for_integrity_error(error):
+        if "account_name_uniq" in str(error).lower():
```
Bewi: Shouldn't it be "name_uniq" ?

js/src/domains/settings/costUnits/components/CostUnitFormWrapper.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        </LoadingButton>
+                    </Stack>
+                </Stack>
+            }
+        >
+            <ExtendedFormikProvider formik={formik}>
+                <CostUnitForm />
+            </ExtendedFormikProvider>
+
+            <Dialog open={confirmOpen} onClose={() => setConfirmOpen(false)}>
```
Bewi: There is a `DeleteModal` in iaso which helps reducing the boiler plate of this.

js/src/domains/settings/costUnits/components/CostUnitList.tsx
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export const CostUnitList: FC<Props> = ({
+    costUnitTypes,
+    onSelectCostUnit,
+    activeCostUnitId,
+}) => {
+    const { formatMessage, formatNumber } = useSafeIntl();
+
+    const getRatioSummary = (costUnit: CostUnitType): string | null => {
```
Bewi: useCallback ?

js/src/domains/settings/costUnits/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const { data: costUnitTypes } = useGetCostUnitTypes();
+
+    useEffect(() => {
+        if (costUnitTypes && costUnitTypes.length > 0 && selected === null) {
+            setSelected(costUnitTypes[0].id);
+        }
+    }, [costUnitTypes, selected]);
+
+    const handleSelectCostUnit = useCallback((costUnit: CostUnitType) => {
```
Bewi: Why not merging `handleSelectCostUnit`, `handleSaved`, `handleDeleted` in one function ?

js/src/domains/settings/costUnits/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (savedId) {
+            setSelected(savedId);
+        }
+    }, []);
+
+    const handleDeleted = useCallback(() => {
+        setSelected(null);
+    }, []);
+
+    const selectedCostUnit =
```
Bewi: Either useMemo or move it to the useEffect which already look for selected change

js/src/domains/settings/costUnits/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+            setSelected(costUnitTypes[0].id);
+        }
+    }, [costUnitTypes, selected]);
+
+    const handleSelectCostUnit = useCallback((costUnit: CostUnitType) => {
+        setSelected(costUnit.id);
+    }, []);
+
+    const handleAdd = useCallback(() => {
+        setSelected('new');
```
Bewi: This is interesting but feels hacky (mixing two concepts :D)
Personal preference is to have another field (isCreating or something) to make it more obvious.
I'll let you decide if you want to keep it short and hacky and a bit more extended.

Bewi (changes requested):
Overall it looks good and works fine, only few comments to address.

Also, I would add some tests for the ratio.

api/cost_unit_types/views.py
michiwend: I agree that `account_name_uniq` might be confusing here, changing it

js/src/domains/settings/costUnits/components/CostUnitFormWrapper.tsx
michiwend: thanks, good catch

Bewi (approved):
LGTM

PR #306, opened 2026-06-10
SNT-510 Move quantity and population to costbreakdown

Bewi:

Move quantity and population to costbreakdown

### Related JIRA tickets

SNT-510

## Changes

Update budget result population and quantity to be part of cost breakdown instead of intervention, org unit or total budget.

## How to test

This data is not showned at the moment, so tests are the only way to verify this.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


michiwend (approved):
LGTM

PR #307, opened 2026-06-10
SNT-511: Fix keys on intervention form

Bewi:

Fix keys on intervention form which ended up seeing duplicate rows

### Related JIRA tickets

SNT-511

## Changes

Properly set keys on intervention form.

## How to test

Before checking out this PR, try to reproduce it by going to **intervention settings** find one with no cost breakdown line in it.
Add two or more lines, save, you should now see twice of them.
Go to another intervention, it'll display those two even though it doesn't belong to that intervention.

Now, checkout this branch and try again, it should work better.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/js/src/domains/settings/interventions/components/InterventionForm.tsx
+++ b/js/src/domains/settings/interventions/components/InterventionForm.tsx
@@ -84,24 +84,28 @@ export const InterventionForm: FC = () => {
             <Stack spacing={2} direction="column">
                 {values.cost_breakdown_lines &&
                     values.cost_breakdown_lines.length > 0 &&
-                    values.cost_breakdown_lines.map((line, index) => (
-                        <InterventionCostBreakdownLineForm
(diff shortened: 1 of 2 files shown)


js/src/domains/settings/interventions/components/InterventionForm.tsx
```diff
@@ -84,24 +84,28 @@ export const InterventionForm: FC = () => {
...
-                                    value,
-                                )
-                            }
-                            onRemove={() =>
-                                removeChildValue('cost_breakdown_lines', index)
-                            }
-                            getErrors={field => getChildError(field, index)}
-                        />
-                    ))}
+                    React.Children.toArray(
```
Bewi, line 87: This is the main issue, when `line.id` was null / undefined.
Using this to have unique keys instead of falling back to index which can give other behavior issues.

michiwend (approved):

PR #310, opened 2026-06-12
Feature SNT-497: Add rule coverage summary and other tweaks

michiwend:

Adds coverage summary widgets to the rule builder

### Related JIRA tickets

SNT-497

## Changes

* Add widgets that show the target population and districts as absolute and percentage values below a rule during edit
* Other tweaks to the rule builder: 
  * Swap name input and color picker
  * Rename sub header "Rule name" -> "Rule name and color"
  * Change rule name input to render the auto placeholder like the actual input text and make it disappear on focus. This way it appears that the auto name is an actual value to the user

## Print screen / video

image


js/src/domains/planning/hooks/usePopulationByOrgUnit.ts
```diff
@@ -0,0 +1,55 @@
+import { useMemo } from 'react';
+import { useGetMetricValues } from '../../dataLayers/hooks/useGetMetrics';
+import { usePlanningContext } from '../contexts/PlanningContext';
+
+// SNT convention: the data layer holding total population values has this
+// MetricType code (see fixtures and demo account seeding).
+const POPULATION_METRIC_CODE = 'POPULATION';
```
Bewi, line 7: I think we should be able to set this up (maybe in account settings)
Some times we will only have custom population layer which means that total population will never be calculated for org units in rules.

js/src/domains/planning/hooks/usePopulationByOrgUnit.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const { data: metricValues } = useGetMetricValues({
+        metricTypeId: populationMetricType?.id,
+    });
+
+    return useMemo(() => {
+        if (!populationMetricType || !metricValues) {
+            return undefined;
+        }
+        const bestYearByOrgUnit = new Map<number, number>();
```
Bewi: Should we use scenario years ?

Bewi (approved):
I think we should have the population  metric code in account settings. 
Just so we can set it up for accounts with custom metrics only.

And the year thing, maybe we should try to get population best year within scenario year range.

Bewi (approved):
LGTM

PR #312, opened 2026-06-16
SNT-520 Fine tune new budget view

Bewi:

Fien tune new budget view

### Related JIRA tickets

SNT-520

## Changes

- Fix alignments, typography
- Add margin on the table
- Header and footer sticky

## How to test

Verify that comments from story are applied to the view.

Note: Footer will not always be on the bottom.

## Print screen / video

image

## Notes

N/A

## Doc

N/A


michiwend (approved):
LGTM

Something we can do later maybe is to have fewer grey tones in one page? But lets merge and move on.

PR #313, opened 2026-06-16
Feature SNT-512: add grant and donor data models and FK to interventions

michiwend:

Donor, Grant models and relations

### Related JIRA tickets

SNT-512, SNT-498

## Changes

Only model changes and migration
* `Donor` model
* `Grant` model
* FKs on `Intervention` and `InterventionAssignment` with `on_delete=SET_NULL`

image


models/grant.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Donor(models.Model):
+    class Meta:
+        app_label = "snt_malaria"
+        ordering = ["name"]
+        constraints = [
+            models.UniqueConstraint(fields=["account", "name"], name="snt_malaria_donor_account_name_uniq"),
+        ]
+
+    account = models.ForeignKey("iaso.Account", on_delete=models.CASCADE, related_name="snt_donors")
```
Bewi, line 12: I wonder if donors should be per account, I guess one donor could be used for multiple account (countries) ?

Bewi (approved):
LGTM, only wondering if `Donor` should be per account or not.

models/grant.py
michiwend: yeah.... but also what about data protection between accounts? We don't know now what will go into the donor model later.

PR #316, opened 2026-06-17
Feature SNT-517: Constrained settings layout with max width and more

michiwend:

* Max width for the settings
* Cancel adding new cost units
* other fixes  

### Related JIRA tickets

SNT-517, SNT-498

## Changes

Settings pages are now width-constrained to 1440px, so the forms don't stretch over a hole screen anymore. Adding of new items have a cancel flow (return to top item in the list). Also the alignment of cost lines has been fixed in the intervention settings.

Max width on a large screen:
image

Cancel adding:
image


Bewi (approved):
LGTM

PR #318, opened 2026-06-17
Feature SNT-518: Add budget settings through "General" tab

michiwend:

Adds a new "General" settings tab and Budget settings under it

### Related JIRA tickets

SNT-518, SNT-498

## Changes

This is only historically related to the grant work but since it was done I'm adding it because it is useful to have budget settings.

image

### DIFF
--- a/js/src/domains/settings/budget/hooks/useBudgetSettingsFormState.tsx
+++ b/js/src/domains/settings/budget/hooks/useBudgetSettingsFormState.tsx
@@ -3,9 +3,12 @@ import { useSafeIntl } from 'bluesquare-components';
 import { FormikHelpers, useFormik } from 'formik';
 import * as Yup from 'yup';
 import { MESSAGES } from '../../../messages';
+import { BudgetSettingsFormValues } from '../types';
 
-const emptyToNull = (value: unknown, originalValue: unknown) =>
-    originalValue === '' ? null : value;
+const emptyToNull = (
+    value: string | null,
+    originalValue: string,
+): string | null => (originalValue === '' ? null : value);
 
 const useValidation = () => {
     const { formatMessage } = useSafeIntl();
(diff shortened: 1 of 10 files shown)


js/src/domains/settings/budget/hooks/useBudgetSettingsFormState.tsx
```diff
@@ -0,0 +1,52 @@
+import { useMemo } from 'react';
+import { useSafeIntl } from 'bluesquare-components';
+import { FormikHelpers, useFormik } from 'formik';
+import * as Yup from 'yup';
+import { MESSAGES } from '../../../messages';
+
+const emptyToNull = (value: unknown, originalValue: unknown) =>
```
Bewi: Isn't there a way to type values and originalValue, I am not that familiar with the `transform` of yup

js/src/domains/settings/budget/hooks/useBudgetSettingsFormState.tsx
```diff
@@ -0,0 +1,52 @@
...
+            }),
+        [formatMessage],
+    );
+};
+
+export const useBudgetSettingsFormState = ({
+    onSubmit,
+    initialValues,
+}: {
+    onSubmit: (values: any, formikHelpers?: FormikHelpers<any>) => void;
```
Bewi: Same here, I would type values and initialValues at least

js/src/domains/settings/general/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                        }
+                    >
+                        <List>
+                            {sections.map(section => {
+                                const Icon = section.icon;
+                                return (
+                                    <ListItem
+                                        key={section.key}
+                                        disableGutters
+                                        disablePadding
```
Bewi: Is the disable padding needed or the sx overrides is enough ?

Bewi (approved):
If you can fix the tipping, would be very appreciate.
Except from that, it looks good.

Bewi (approved):
LGTM

PR #320, opened 2026-06-18
[IA-5219] prepare plugin for docker + new CI

tdethier:

use new workflow for PRs

### Related JIRA tickets

IA-5227, IA-5219

## Changes

- CI:
  - added new workflows that call `iaso-ci` ([dedicated repo](https://github.com/BLSQ/iaso-ci), `uv` for backend, building docker image, deploying with docker...)
  - deleted old `ci.yml` 
  - renamed `deploy` as `old_deploy`
 
- changes to frontend to make eslint happy
- added `docker-compose.prod.yml` for deployments
- added missing BE translations

## How to test

- check CI
- the SNT image was deployed and tested on iaso-upgrade

## Print screen / video

/

## Notes

/

## Doc


.github/workflows/develop.yml
```diff
@@ -0,0 +1,56 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  config:
+    uses: BLSQ/iaso-ci/.github/workflows/config.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
```
Bewi: I guess we'll need to update those uses to target develop

.github/workflows/docker_deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+
+
+on:
+  workflow_dispatch:
+    inputs:
+      # Branch is chosen by default in github manual actions
+      environment:
+        description: Env (snt-malaria, snt-malaria-staging, snt-malaria-demo) or exact server name (iaso-snt-malaria-staging-worker, iaso-snt-malaria, iaso-snt-malaria-demo-worker)
+        required: true
+        default: snt-malaria
```
Bewi: I wonder if we shouldn't change this to be snt-malaria-staging to prevent deploying to prod if someone forgot to change this.

.github/workflows/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  config:
+    uses: BLSQ/iaso-ci/.github/workflows/config.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
```
Bewi: Same here, I guess we should change the branch

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,43 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  config:
+    uses: BLSQ/iaso-ci/.github/workflows/config.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
```
Bewi: Same in this file, we should change the branch name

Bewi (changes requested):

.github/workflows/docker_deploy.yml
tdethier: yeah I was wondering the same, usually staging is the default but on SNT the default in the old deploy CI was prod, so I left it like that

tdethier: but i agree, let's switch this back to develop/staging

Bewi (approved):
LGTM

PR #322, opened 2026-06-22
SNT-521 Add line item calculation hint

Bewi:

Add line item tooltip on budget view

### Related JIRA tickets

SNT-521

## Changes

- Add buffer to budget settings
- When calculating budget, also save information about cost units, cost line. 
- Display a tooltip on costline of budget view with Target population, Unit cost, Ratio, Buffer.

## How to test

Go to snt, make sure budget is refreshed by modifying one of the coverage value. 
This should work.
Then check the (i) icon and verify that information displayed matches your config.
Note for the unit ratio, if it is inverted, label should be `Poeple per {unit}` if default, if should be `{unit} per person`.

## Print screen / video

image

## Notes

N/A

## Doc


michiwend (commented):
I'm a bit confused by the second line, what is it supposed to tell me?
In the settings it is "one Bale covers 90 people"

Image

services/budget/budget_calculation.py
```diff
@@ -425,6 +457,13 @@ def _build_org_units_costs(
             )
         return org_units_costs
 
+    def _get_ratio(self, unit_type):
+        ratio = unit_type.ratio if unit_type else None
+        if ratio:
+            ratio = 1 / ratio if unit_type.invert_value else ratio
```
michiwend: this is a double inversion because ratio is already a model property that does that

js/src/domains/messages.ts
```diff
@@ -90,6 +90,30 @@ export const MESSAGES = defineMessages({
...
+        defaultMessage: 'Conversion factor',
+    },
+    budgetingCostLinePeoplePerUnit: {
+        id: 'iaso.snt_malaria.budgeting.costLine.peoplePerUnit',
+        defaultMessage: 'People per {unit}',
+    },
+    budgetingCostLineUnitPerPeople: {
+        id: '',
+        defaultMessage: '{unit} per person',
+    },
```
michiwend, line 112: no `id` and translations for this one

michiwend (changes requested):

services/budget/budget_calculation.py
Bewi: I am now using the value instead

michiwend (approved):

PR #324, opened 2026-06-23
Feature SNT-495: Scenario Summary

michiwend:

New Scenario Summary page with Grant tracking

### Related JIRA tickets

SNT-495

## Changes

- Refactoring
  - move `../domains/compareCustomize/Card` to the shared `/components/WidgetCard` component
  - replace generic action slot in `WidgetCard` with a dropdown prop, use small MUI select
  - Define shared styles and helpers for consistent charts across the app
- Initial version of the new Summary Page with four widgets
  - New cost-by-grant chart which integrates with previous backend work on grant tracking
  - Bring back "old" cost-by-district map
  - Prevalence widget using impact data, if available
  - New cost-per-intervention breakdown chart
  - Show summary page by default for locked scenarios

## How to test

* Make sure you have working impact or use the fake provider to get prevalence data
* Set up at least one grant in the settings with a certain envelope
* Assign the grant to at least one intervention in the settings, make sure the intervention has at least one cost item
* Create or update a scenario and make sure the intervention is assigned and produces a non-zero cost line
* Check the new summary page, you should get a plot for the total costs and the cost for the grant against the configured grant envelope. Compare the totals with the budget page
* Lock the scenario, return home, enter the scenario. You should see the summary page right away

## Print screen / video

image


js/src/domains/planning/index.tsx
```diff
@@ -210,42 +304,52 @@ export const Planning: FC = () => {
...
-                                        />
-                                    )}
-                                    {activeTab === 'budget' &&
-                                        orgUnits &&
-                                        budget && <BudgetTable />}
-                                </CardStyled>
-                            </Card>
-                        </PaperFullHeight>
+                        {activeTab === 'summary' ? (
+                            <PaperFullHeight sx={styles.summaryColumn}>
```
Bewi: Could move this to own component to not poluate to much here ?

js/src/domains/planning/index.tsx
```diff
@@ -65,6 +135,17 @@ export const Planning: FC = () => {
     const { formatMessage } = useSafeIntl();
     const [activeTab, setActiveTab] = useState('map');
 
+    // Locked scenarios default to the summary tab, once, on first load.
+    const didInitActiveTab = useRef(false);
```
Bewi, line 95: Do we actually need this ?
Feels like the useEffect is only refreshed if scenario changes in which case we do want to refresh active tab.
I guess it is to cover the case when you lock the scenario, to not switch tab ?

js/src/domains/planning/libs/budget-aggregation.ts
```diff
@@ -0,0 +1,89 @@
...
+        (a, b) => ({ ...a, total_cost: a.total_cost + b.total_cost }),
+    );
+}
+
+function mergeInterventions(
+    interventions: BudgetIntervention[],
+): BudgetIntervention[] {
+    return mergeByKey(
+        interventions,
+        intervention => `${intervention.type} - ${intervention.code}`,
```
Bewi: Why not using the intervention id ?
We made quite an effort to get out of the type / code keys.

js/src/domains/planning/components/budgeting/BudgetSummary.tsx
```diff
@@ -0,0 +1,320 @@
...
+};
+
+// Custom bar shape: a single `<Bar dataKey="cost">` is rendered so recharts
+// sizes (x, y, width, height) the bar against the cost value. We derive a
+// pixels-per-value factor from that sizing and redraw the bar as:
+//  - a grey rect up to min(cost, amount),
+//  - a red rect for the part of cost that exceeds the grant amount,
+//  - a dashed, unfilled rect outlining the grant amount target.
+// Only the topmost segment gets rounded top corners.
+const BudgetBarShape: FC<BudgetBarShapeProps> = ({
```
Bewi: Move this to own component

js/src/domains/planning/components/budgeting/BudgetSummary.tsx
```diff
@@ -0,0 +1,320 @@
...
+        `L${x + width - r},${y}`,
+        `Q${x + width},${y} ${x + width},${y + r}`,
+        `L${x + width},${y + height}`,
+        'Z',
+    ].join(' ');
+};
+
+// Same shape but left open at the bottom, so the envelope outline has no line
+// along the baseline.
+const roundedTopOutlinePath = (
```
Bewi: Parts of me want this to be moved to some utils

js/src/domains/planning/components/budgeting/CostPerInterventionSummary.tsx
```diff
@@ -0,0 +1,375 @@
...
+    index: number,
+    count: number,
+): string => {
+    if (count <= 1) return base;
+    return lighten(base, (index / (count - 1)) * MAX_LIGHTEN);
+};
+
+// Rectangle with only its right corners rounded, so the outer end of the bar
+// is rounded like the design while it stays flush with the value axis.
+const roundedRightRectPath = (
```
Bewi: Same here, we could move this to utils

Bewi (changes requested):

Bewi (approved):
LGTM

PR #325, opened 2026-06-23
SNT-426 Only show trash icon on hover and remove edit button

Bewi:

Only show trash icon on hover and remove edit button

### Related JIRA tickets

SNT-526

## Changes

Rule list: 
- remove edit button and handle rule box click.
- Show delete button on hover of a rule box.

## How to test

Go to SNT, Scenario, in rule list, no action should be visible by default.
By hovering it, delete button should be shown and working.
By clicking on a rule, it should open the edit form.
Test the rule name overflow too.

## Print screen / video

image

image


michiwend (approved):
There is a minor bug that when you dismiss the dialog with "No", the bin remains visible on the card. Otherwise LGTM

michiwend:
Another knit pick. The area for the bin to show up and the pointer to change to "pointy" is not the full box but the inner container. So when you touch it only outside, it doesn't show the bin and the other pointer. Maybe we can put this on the outer box

PR #327, opened 2026-06-25
SNT-526 add buffer to general settings

Bewi:

Add buffer to general settings

### Related JIRA tickets

SNT-526

## Changes

Add buffer to general settings tab.
Add a warning to let user know that budget are not auto matically refreshed.

## How to test

Go to settings, general tab and modify the buffer.
Go to a scenario with a budget, save a rule or change a coverage to recalculate the budget and verify on the tooltip that it is using your new value.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


michiwend (approved):
LGTM

PR #328, opened 2026-06-25
SNT-525 introduce default population setting

Bewi:

Introduce default population setting

### Related JIRA tickets

SNT-525

## Changes

- Add a new property default_population to account settings
- Add a view to manage account settings
- use default_default population on rule to show target population count

## How to test

Go to snt, settings, play arround with the account settings view.
Only org units types from your account should be shown.
Only population metrics should be listed in default population field.

Go to a rule add / edit and check that if you change the default_population, the count on the bottom changes as well.

## Print screen / video

image

image

## Notes

N/A

## Doc

N/A


js/src/constants/translations/en.json
```diff
@@ -348,5 +348,11 @@
...
     "iaso.snt_malaria.settings.budget.invalidCurrencyCode": "Enter a 3-letter currency code.",
     "iaso.snt_malaria.settings.budget.buffer": "Buffer multiplier",
-    "iaso.snt_malaria.settings.budget.bufferHelp": "Multiplier applied to all costs, e.g. 1.1 for a 10% buffer."
+    "iaso.snt_malaria.settings.budget.bufferHelp": "Multiplier applied to all costs, e.g. 1.1 for a 10% buffer.",
+    "iaso.snt_malaria.settings.account.title": "Account settings",
+    "iaso.snt_malaria.settings.account.subtitle": "Configure the org unit types and default population metric used across the account.",
+    "iaso.snt_malaria.settings.account.interventionOrgUnitType": "Intervention org unit type",
+    "iaso.snt_malaria.settings.account.focusOrgUnitType": "Focus org unit type",
+    "iaso.snt_malaria.settings.account.defaultPopulation": "Default population",
+    "iaso.snt_malaria.settings.account.defaultPopulationHelp": "Metric type used to display population totals on the rule panel."
```
michiwend: I wouldn't scope this so much down. I think we could say this in a more general way that this is the default poplation metric used across the app wherever population is used.

michiwend (changes requested):
Mostly LGTM but I would also use this as the default for driven cost items and change the wording (see below)

michiwend (approved):

PR #329, opened 2026-06-30
SNT-531 Multi map support in data layer

Bewi:

Multi map support in data layer

### Related JIRA tickets

SNT-531

## Changes

- Add data layer comparis context
- Add an action to add a layer to the comparison list
- Add styles for small maps
- Add an action to remove layer from the comparison list

## How to test

Go to data layer, hover one of them, you should see a "+" button. 
Clicking on the line, should select the layer as main layer.
Clicking on the "+" should only add it to the comparison list.
Comparison list should be smaller.
No org unit selection should be available on small maps.
Verify that you can add a maximum of 4 map to the comparison panel.
Try to add twice the same map, then removing one should remove the one in the correct index.

## Print screen / video

https://github.com/user-attachments/assets/655718f5-452f-4e12-9cfb-de4d0277deaa

## Notes

N/A

## Doc

### DIFF
--- a/js/src/domains/dataLayers/contexts/DataLayerComparisonContext.tsx
+++ b/js/src/domains/dataLayers/contexts/DataLayerComparisonContext.tsx
@@ -1,0 +1,78 @@
+import React, {
+    createContext,
+    useCallback,
+    useContext,
+    useState,
+} from 'react';
+import { OrgUnit } from 'Iaso/domains/orgUnits/types/orgUnit';
+import { MetricType } from '../types/metrics';
+
+const MAX_METRICS = 4;
+
+type DataLayerComparisonContextType = {
+    maxMetricsCountReached: boolean;
+    comparisonMetricTypes: MetricType[];
+    addMetricToComparison: (metricType: MetricType) => void;
+    // We need the index here as we can have multiple time the same metric on state.
+    removeMetricFromComparison: (metricType: number, index: number) => void;
+    orgUnits: OrgUnit[];
+};
+
+const DataLayerComparisonContext =
+    createContext<DataLayerComparisonContextType>({
+        maxMetricsCountReached: false,
+        comparisonMetricTypes: [],
+        orgUnits: [],
+        addMetricToComparison: () => {},
+        removeMetricFromComparison: () => {},
+    });
+
+export const useDataLayerComparisonContext = () =>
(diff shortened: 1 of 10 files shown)


michiwend:
A few UI findings:

- Would harmonise the widget gap with rest of the app (summary page, compare & customize):
   image image
- There is a glitch where it shows the add button briefly and then it disappears again. It happens, when you enter a list entry at the position where the button is. When you hover it at a different spot and then move over, it works: 
   image
- IMO when adding two breakout maps, they should be exactly 100% of the main map. Right now they are a little higher and the bottom one overflows the container:
   image

js/src/domains/dataLayers/contexts/DataLayerComparisonContext.tsx
```diff
@@ -0,0 +1,78 @@
+import React, {
+    createContext,
+    useCallback,
+    useContext,
+    useState,
+} from 'react';
+import { OrgUnit } from 'Iaso/domains/orgUnits/types/orgUnit';
+import { MetricType } from '../types/metrics';
+
+const MAX_METRICS = 4;
```
michiwend, line 10: Since you made it scrollable, why do we need to limit it?

michiwend (changes requested):
Impl. LGTM. Have a few UI remarks below

js/src/domains/dataLayers/contexts/DataLayerComparisonContext.tsx
Bewi: As we can have multiple time a map for the same metric, this is to prevent them adding to many maps which would eventually cost to much on performances

Bewi:
> A few UI findings:
> 
>     * Would harmonise the widget gap with rest of the app (summary page, compare & customize):
>       image image
> 
>     * There is a glitch where it shows the add button briefly and then it disappears again. It happens, when you enter a list entry at the position where the button is. When you hover it at a different spot and then move over, it works:
>       image
> 
>     * IMO when adding two breakout maps, they should be exactly 100% of the main map. Right now they are a little higher and the bottom one overflows the container:
>       image

Fixed spacing / sizing.
For the glitch, this is coming from the tooltip which is a material UI component, there isn't a real fix for this except moving it further which then looks weird. So I will leave it like this now.

michiwend (approved):

PR #330, opened 2026-06-30
Feature SNT-519: make cost unit driver explicit

michiwend:

Explicit cost unit driver

### Related JIRA tickets

SNT-519

## Changes

Add an `is_proportional` flag to cost units. Proportional units carry a conversion factor and drive cost lines by a population metric; the target population becomes required on those lines. Non-proportional units are absolute/fixed cost: their ratio is always 1 (regardless of the previous value when switching) and cost lines using them have no cost driver.

Surface the cost driver in the scenario cost breakdown with an icon (MUI _groups_ for proportional, _numbers_ for fixed) and consistent tooltips on the rows and the yearly inputs.

## How to test

- Set a unit to fixed
- Check that the cost line doesn't require a population layer (and doesn't render it)
- Set a unit to Proportional
- Check that the cost line requires a population layer
- Seed a demo account
- Check that `Each` and `Days` are set to fixed

## Print screen / video

Checkbox in Unit type toggles conversion factor and direction
image
image

Cost items don't show driver dropdown for fixed units and require it for driven ones
image

Cost breakdown features icons and tooltips indicating the difference
image


Bewi (changes requested):
We could get rid of the `fixed_cost`, it feels duplicating the `is_proportion`.
Also, on small screen, the icon aside of the label create some overflow even for short names, I wonder if it couldn't be moved a bit more on the left ?

Image

image

Bewi (approved):
LGTM

PR #337, opened 2026-07-10
Feature SNT-530: node editor improvements

michiwend:

First round of improvements to the node editor

### Related JIRA tickets

SNT-530

## Changes

* Info texts in the transformation nodes
* New node type _Combine_: Combines dynamic amount of inputs using either of `min`, `max`, `sum`, `mean`
* New node type _Normalise_: Scales the value-range of all org units to a new range `0-1` or -`0 - 100`

## Print screen / video

image

image

image


Bewi (approved):
LGTM

PR #338, opened 2026-07-10
SNT-533 generate flume by claude

Bewi:

Introduce a ChatPanel and generate Flume by claude

### Related JIRA tickets

SNT-533

## Changes

- Add a chat panel (only visible if there is an ai key setup for the account)
- Generate flume config by Claude
- Apply flume config from Claude

## How to test

First make sure you have an Anthropic API key set on your account.
Then go to DataLayer and create a new composite layer.
Ask some stuff (I.E. Give me prevalence based on WHO standards) to the chat and verify that flume diagram is generated properly.

## Print screen / video

image

## Notes

## Doc

### DIFF
--- a/js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
+++ b/js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
@@ -1,0 +1,15 @@
+import { postRequest } from 'Iaso/libs/Api';
+import { useSnackMutation } from 'Iaso/libs/apiHooks';
+import { CompositeLayerAIRequest, CompositeLayerAIResponse } from './types';
+
+export const useSendCompositeLayerAIMessage = () => {
+    return useSnackMutation<
+        CompositeLayerAIResponse,
+        Error,
+        CompositeLayerAIRequest
+    >({
+        mutationFn: (data: CompositeLayerAIRequest) =>
+            postRequest('/api/snt_malaria/composite_layer_ai/', data),
+        showSuccessSnackBar: false,
+    });
+};
--- a/js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
+++ b/js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
@@ -1,15 +1,13 @@
+import { useMutation } from 'react-query';
 import { postRequest } from 'Iaso/libs/Api';
-import { useSnackMutation } from 'Iaso/libs/apiHooks';
 import { CompositeLayerAIRequest, CompositeLayerAIResponse } from './types';
 
 export const useSendCompositeLayerAIMessage = () => {
-    return useSnackMutation<
+    return useMutation<
         CompositeLayerAIResponse,
         Error,
         CompositeLayerAIRequest
-    >({
-        mutationFn: (data: CompositeLayerAIRequest) =>
-            postRequest('/api/snt_malaria/composite_layer_ai/', data),
-        showSuccessSnackBar: false,
-    });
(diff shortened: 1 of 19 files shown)


plugin_settings.py
```diff
@@ -10,6 +10,7 @@
     # Paths that IasoLogoutView accepts as ?next=... after logging the user out.
     # Used by the configureAccount wizard's "Restart" link.
     "LOGOUT_NEXT_ALLOWED_PATHS": ["/snt_malaria/public/setupAccount"],
+    "COMPOSITE_LAYER_AI_MODEL": os.environ.get("COMPOSITE_LAYER_AI_MODEL", "claude-opus-4-7"),
```
tdethier, line 13: @Bewi don't forget to include this in the SNT docker compose file

tests/api/composite_layer_ai/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+            ),
+            "conversation_history": [
+                {"role": "user", "content": "Create a composite layer"},
+                {"role": "assistant", "content": "Here is your composite layer."},
+            ],
+        }
+
+    def test_unauthenticated_returns_401(self):
+        response = self.client.post(BASE_URL, {"message": "hi"}, format="json")
+        self.assertEqual(response.status_code, 401)
```
tdethier: nitpick: using const values from `rest_framwork.status` is a bit more clear

tests/api/composite_layer_ai/test_views.py
```diff
@@ -0,0 +1,151 @@
...
+        super().setUp()
+
+        self.account = Account.objects.create(
+            name="Test Account",
+            anthropic_api_key="sk-test-key",
+        )
+        self.account_no_key = Account.objects.create(name="Account Without Key")
+
+        self.user = self.create_user_with_profile(
+            username="user_with_key", account=self.account, permissions=[SNT_SETTINGS_WRITE_PERMISSION]
```
tdethier, line 28: i would add a test that checks for user permissions

js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const useSendCompositeLayerAIMessage = () => {
+    return useSnackMutation<
+        CompositeLayerAIResponse,
+        Error,
+        CompositeLayerAIRequest
+    >({
+        mutationFn: (data: CompositeLayerAIRequest) =>
+            postRequest('/api/snt_malaria/composite_layer_ai/', data),
+        showSuccessSnackBar: false,
```
michiwend: `showSuccessSnackBar` only silences the success toast... `useSnackMutation`'s default `onError` still fires a generic snackbar on every failure, on top of the friendly error bubble, so users see two different error messages for one failure

 consider `ignoreErrorCodes` for this endpoint's known codes

michiwend (approved):
LGTM with one pick

tdethier (changes requested):
Overall the feature works fine, I managed to create several composite layers 👍 

I'm requesting changes because:
- the new env var needs to be added to the SNT docker compose file
- I cannot edit my composite layers, it looks like they are not recognized as composite - update: it was because the dev feature flag was missing

----------------------


editing a composite layer and asking AI to add a new layer to the composite completely deletes the previous work, which is annoying
image

js/src/domains/compositeLayerEditor/compositeLayerChatBot/useSendCompositeLayerAIMessage.ts
Bewi: Fixed this by using the useMutation instead of useSnackMutation

tdethier (approved):
LGTM 👍

PR #339, opened 2026-07-16
Feature SNT-538: Move conversion to cost items

michiwend:

Make cost units "fundamental" and move cost conversion to cost items

### Related JIRA tickets

SNT-538

## Changes

This change moves all the cost conversion logic from cost units into cost items. The fact that it was on the units only led to confusion and further prevented us from implementing commodity tracking per fundamental unit. For example, for SMC we had three different units to account for the cycles but that prevented us from tracking the total amount of SMC doses.

* Move conversion factor and direction to cost items
* Units are simple "labels" now with only a description
* Units are consolidated as we don't need different units for e.g. different SMC cycles etc
* New cost item interface is explicit about fixed vs proportional cost items
* Database migrations migrate existing units and interventions to new, consolidated ones
* Seeder scripts updated

## How to test

Test with existing data
* Before you pull and run, take some screenshots of existing local scenarios
* Run the PR locally
* Check that units were consolidated and interventions were updated. Cost items that were using proportional units before should now be proportional cost items and carry the factor and direction that was previously on the unit.
* Compare your scenario's cost with your screenshots

Compare demo accounts
* Run the demo seeder and recreate the demo account locally
* Compare the two demo scenarios' costs with the ones from our official demo account

## Print screen / video

New Interface for setting up conversion per cost item
image

Consolidated cost unit types
image


migrations/0055_move_conversion_to_cost_lines.py
```diff
@@ -0,0 +1,153 @@
...
+BUDGET_RESULT_KEY_RENAMES = {
+    "cost_unit_ratio": "conversion_factor",
+    "cost_unit_inverted": "invert_conversion_factor",
+}
+
+
+def copy_unit_conversion_to_cost_lines(apps, schema_editor):
+    CostUnitType = apps.get_model("snt_malaria", "CostUnitType")
+    InterventionCostBreakdownLine = apps.get_model("snt_malaria", "InterventionCostBreakdownLine")
+    db_alias = schema_editor.connection.alias
```
tdethier, line 43: why do you need an alias?

models/cost_breakdown.py
```diff
@@ -46,6 +56,15 @@ class Meta:
...
-        return self.population_layer_id is None
+    def conversion_ratio(self):
+        """Canonical conversion factor used by the budget calculation (1 / value when inverted).
+
+        Fixed cost lines always return 1, regardless of the stored value.
+        """
+        if not self.is_proportional or self.conversion_factor is None:
+            return Decimal(1)
+        if self.invert_conversion_factor:
+            if self.conversion_factor == 0:
```
tdethier: I guess you could include this `if == 0` in the if above on line `64`

tests/models/test_cost_breakdown_line.py
```diff
@@ -0,0 +1,43 @@
+from decimal import Decimal
+
+from plugins.snt_malaria.models import InterventionCostBreakdownLine
+from plugins.snt_malaria.models.cost_unit_type import CostUnitType
+from plugins.snt_malaria.tests.common_base import SNTMalariaTestCase
+
+
+class InterventionCostBreakdownLineModelTestCase(SNTMalariaTestCase):
```
tdethier, line 8: nitpick: you could add a test where `conversion_factor` is `None` when calling `conversion_ratio()`

services/budget/budget_calculation.py
```diff
@@ -338,8 +336,7 @@ def _compute_fixed_cost_row(self, line, year, inflation_multiplier, intervention
         Calculate using yearly value as quantity. Added once per intervention regardless of org units.
         """
         yearly_value = self._get_yearly_value(line, year)
-        unit_ratio = line.unit_type.ratio if line.unit_type and line.unit_type.ratio is not None else Decimal("1")
-        quantity = yearly_value * Decimal(str(unit_ratio))
+        quantity = yearly_value
```
tdethier: i guess you could do directly `quantity = self._get_yearly_value(line, year)`

api/intervention_cost_breakdown_line/serializers.py
```diff
@@ -145,12 +153,16 @@ def get_fields(self):
...
     def validate(self, attrs):
         attrs = super().validate(attrs)
-        unit_type = attrs.get("unit_type")
-        if unit_type is not None and not unit_type.is_proportional:
+        if attrs.get("is_proportional"):
+            if not attrs.get("population_layer"):
+                raise serializers.ValidationError(
+                    {"population_layer": "A target population is required for proportional cost items."}
+                )
+            if attrs.get("invert_conversion_factor") and attrs.get("conversion_factor") == 0:
```
tdethier: should this if be placed at the same level as `if attrs.get("is_proportional"):`? you also want to check that if the object has `is_proportional = False`, right?

js/src/utils/pluralize.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Naive pluralization for user-provided labels (e.g. cost unit names):
+ * appends an "s" unless the value is exactly 1.
+ */
+export const pluralize = (label: string, value: number): string =>
```
tdethier, line 6: isn't there a generic lib for that? because what happens when `value == 0`, or if the word ends with `s`? 😅

js/src/domains/planning/components/budgeting/CostLineRow.tsx
```diff
@@ -283,9 +284,17 @@ export const CostLineTooltip: FC<TooltipProps> = ({ costLine }) => {
...
+                          costLine.unitName,
+                          costLine.conversionFactor ?? 1,
+                      ),
                   })),
-        [costLine.unitName, costLine.invertedConversionFactor, formatMessage],
+        [
+            costLine.unitName,
+            costLine.invertedConversionFactor,
+            costLine.conversionFactor,
+            formatMessage,
```
tdethier, line 296: shouldn't `pluralize()` be in the array deps as well?

js/src/domains/settings/interventions/components/InterventionCostBreakdownLineForm.tsx
```diff
@@ -21,112 +31,210 @@ export const InterventionCostBreakdownLineForm: FC<Props> = ({
...
+                }),
+            },
+            {
+                value: 'inverse',
+                label: formatMessage(MESSAGES.budgetingCostLinePeoplePerUnit, {
+                    unit: selectedUnitLabel,
+                }),
+            },
+        ];
+    }, [formatMessage, selectedUnitLabel, costBreakdownLine.conversion_factor]);
```
tdethier, line 64: same comment about `pluralize()` in array deps

tdethier (approved):
Code looks good overall 👍 
Budget values on the demo account are the same before and after this PR.

I left a few comments here and there

PR #340, opened 2026-07-16
Feature SNT-527: Unit-based commodity tracking

michiwend:

Track commodities per unit procured

### Related JIRA tickets

SNT-527

## Changes

* Include the buffer in the quantities, not only in the costs
* Specify units to be commodities with a checkbox
* Detailed unit quantities in the budget page
* Commodity summary widgets on the summary page (only procurement and the new flag)

## Testing

Similar to #339 we should compare the budget calculation to a known output because it touches the calculation (buffer into quantity), so:

* Recreate local demo account and compare with our demo env that all calculations produce the same costs

## Print screen / video

Flag a unit to be included in the commodities tracking
image

Unit quantity breakdown in the budget page
image

Commodities summary
image


tdethier (approved):
looks good, works well 👍 

nitpick: the area becomes a bit crowded when there are multiple commodities, having multiple cells in the middle of all the charts looks off

Image

PR #343, opened 2026-07-22
[IA-5329] fix SNT cron

tdethier:

fix SNT cron

### Related JIRA tickets

IA-5329

## Changes

- fix cron after switch to docker
- updated "manual" cron into an EBS cron that triggers a task (standard procedure in vanilla iaso)

## How to test

- can't be tested locally, this needs to be deployed on a server - to deploy this change, you need this [iaso-ci PR](https://github.com/BLSQ/iaso-ci/pull/3) to be merged
- when deployed: check cron logs, check the expected result of the cron job 

## Print screen / video

/

## Notes

/

## Doc

/


tasks/recreate_demo_account.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def log_and_report_failure(task: Task, message: str, error: Exception):
+    logger.exception(message)
+    task.report_failure(error)
+
+
+@task_decorator(task_name="recreate_demo_account")
+def recreate_demo_account(task=None):
+    if settings.ENVIRONMENT != "staging":
```
Bewi: Are we sure about this ?
I guess environment should be changed to "demo" ?

Bewi (changes requested):

tasks/recreate_demo_account.py
tdethier: yeah i'm going to remove this, it was just some setup to test the feature on staging 😅

i'll revert all the setup commits

Bewi (approved):
LGTM

PR #347, opened 2026-07-24
SNT-537 Manage interventions

Bewi:

Manage interventions and intervention categories

### Related JIRA tickets

SNT-537

## Changes

- CRUD on intervention and intervention categories
- Forms for intervention and intervention categories

## How to test

Go to settings and play around: 

- Remove soft deletion from intervention and intervention categories
- Add an intervention to an existing category
- Add an intervention category
- Add an intervention to that new category
- Delete that intervention (it should work as no rule are linked to it)
- Add an intervention with the same name, it should work
- Add an intervention with a name already taken, it should fail
- Delete a category, it should fail if an intervention is linked to it
- Go to Scenarios and add a rule for a custom intervention, it should work and calculate the budget as any other intervention
- Now try to delete that intervention you linked to a rule, it should fail

## Print screen / video

image

image

## Notes

N/A

## Doc

N/A


api/interventions/views.py
```diff
@@ -9,20 +11,47 @@
...
+    def perform_destroy(self, instance):
+        # Intervention has no DB-level protection against deletion while a scenario
+        # rule references it (the FK lives on the rule's M2M through table), so block
+        # the delete ourselves while any scenario rule still uses this intervention.
+        if ScenarioRule.objects.filter(interventions=instance).exists():
+            raise MethodNotAllowed(
+                self.request.method,
+                _("Cannot delete this intervention because it is used by one or more scenario rules."),
+            )
+        instance.delete()
```
michiwend: I wonder if that is a bit too loose of a check and too risky.
This checks only for assigned rules but not for generated assignments or cost data which cascade, too. Also we could maybe do a migration and have a real db-level guard?

api/intervention_categories/views.py
```diff
@@ -9,9 +14,44 @@
...
+                _("Cannot delete this intervention category because it is used by one or more interventions."),
+            )
+
+    @staticmethod
+    def _raise_for_integrity_error(error):
+        if "name_uniq" in str(error):
+            raise serializers.ValidationError(
+                {"name": _("An intervention category with this name already exists.")}
+            )
+        raise serializers.ValidationError(str(error))
```
michiwend, line 55: this surfaces a raw db error

js/src/domains/settings/interventions/components/InterventionFormWrapper.tsx
```diff
@@ -61,23 +138,39 @@ export const InterventionFormWrapper: FC<Props> = ({ interventionId }) => {
...
         });
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [interventionDetails, metricTypes]); // Only run when interventionDetails or metricTypes changes
 
+    const handleSave = useCallback(async () => {
+        await basicFormik.submitForm();
+        if (!isNew) {
+            await detailsFormik.submitForm();
+        }
+    }, [basicFormik, detailsFormik, isNew]);
```
michiwend: both forms share one save button, so editing an existing intervention, clearing the required code field and clicking save will show a validation error under "Code" (basic form didn't save) but will still fire the `PATCH` for the details form

michiwend (changes requested):

api/interventions/views.py
Bewi: Yeah, I was curious about assignments, but in theory there should not be any if there are no rule matching.
And blocking it if there are trailing assignments could block the user.
As for cost, for me it is fine to cascade it, it is only attached to one intervention anyway.
What do you think ?

michiwend: We can cascade it all the way but maybe let the user know about the consequence. Maybe even something like "This interventions is still linked to xxx cost lines and yyy assignments, if you proceed they get deleted, too"

michiwend (approved):

PR #349, opened 2026-07-28
SNT-542: Enable data import for years starting 2000

Bewi:

Extend year range when importing layers

### Related JIRA tickets

SNT-542

## Changes

N/A

## How to test

Go to snt, data layers and import for 2020, verify that import worked properly

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


michiwend (changes requested):
This is a big range but okay if they want it. Only thing I would change is, that the default should be "now" instead of `yearOptions[0].value` which would be the earliest year. Also the sort order of the dropdown list I would go from newest to oldest

michiwend (approved):

PR #350, opened 2026-07-28
Feature SNT-552: New create/edit flow for composite layers

michiwend:

New create/edit flow for composite layers utilising the existing create/edit layer dialogue 

### Related JIRA tickets

SNT-552
SNT-546

## Changes

* Remove the extra "Create composite layer" button
* Remove the extra "Edit composite layer" context menu entry
* Integrate composite layer creation into the existing "Create layer" flow by offering a layer type selector
* Edit existing composite layers threw a new button on the top right of the view

## Print screen / video
image

image

image


js/src/domains/dataLayers/dataLayerForm/DataLayerDialog.tsx
```diff
@@ -75,21 +123,78 @@ export const DataLayerDialog: FC<MetricTypeDialogProps> = ({
...
                 ),
             };
         }
         return undefined;
-    }, [metricType]);
+    }, [metricType, isEditingComposite, compositeLayer]);
 
     const onSubmit = (values: MetricTypeFormModel) => {
+        // Composite variant: name/legend are owned here, the values themselves come from a graph.
+        if (values.is_composite) {
```
Bewi: Feels like the onSubmit could be splitted in two

js/src/domains/dataLayers/dataLayerForm/DataLayerForm.tsx
```diff
@@ -46,26 +54,134 @@ export const MetricTypeForm: FC<MetricTypeFormProps> = ({
...
+                label: formatMessage(COMPOSITE_MESSAGES.legendOrdinal),
+            },
+        ],
+        [formatMessage],
+    );
+
+    // Composite is offered only when available, and never as a conversion of an existing layer
+    // (a composite owns a graph, so the variant is fixed at creation).
+    const layerType = isComposite ? 'composite' : 'data';
+    const layerTypeOptions = useMemo(() => {
```
Bewi: I would move this to the layerTypeSelect component

js/src/domains/dataLayers/dataLayerForm/LayerTypeSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+type Props = {
+    label: string;
+    value: string;
+    options: LayerTypeOption[];
+    onChange: (value: string) => void;
+    disabled?: boolean;
+};
+
+/**
+ * Outlined select whose options carry a leading icon — the shared InputComponent select can't.
```
Bewi: Unnecessary comments

js/src/domains/dataLayers/dataLayerForm/LayerTypeSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+    options,
+    onChange,
+    disabled = false,
+}) => (
+    <FormControl
+        fullWidth
+        size="small"
+        disabled={disabled}
+        sx={{
+            '& .MuiSelect-select': {
```
Bewi: Move the sx to a styles object that satisfies SxStyles

js/src/domains/dataLayers/dataLayerForm/DataLayerForm.tsx
Bewi: Or make the layerTypeSelect a reusable component such as : 
IconSelectComponent or smth

js/src/domains/dataLayers/dataLayerMap/DataLayerMapWrapper.tsx
```diff
@@ -40,6 +43,11 @@ type Props = {
     displayOrgUnitId?: number;
     small?: boolean;
     onRemove?: () => void;
+    /** Whether composite layers are available (dev-features flag + settings-write permission). */
```
Bewi: No fan of those comments, this input doesn't actually care of why we show/hide it just does it :D

Bewi (changes requested):
Overall, I find it a bit confusing. 

I think the only thing shared between node editor and metadata is the legend.
And there is a lot of logic to keep them sync.

First thing would be to have a patch request on Data layers (metric types), which doesn't care about what it receives and why, it only updates given fields.

Then I wonder if we couldn't just create a metric type, not telling if it is composite or not at first.
Then you can decide, either you import data for it, either you add a composite layer.
Based on that, we could define if it is a composite or a normal layer.
This way, we reduce the role of composite layer endpoint.
Plus, we can get rid of all the draft logic.

From then, composite layer endpoint will only be responsible of its own domain.
The only thing left is the legend and I guess that if it is only that, it should be fine.

Bewi (approved):
LGTM

PR #351, opened 2026-07-28
Fix SNT-553: Node looses connector on click

michiwend:

Fix, when connected node input ports are clicked, removes the connector for real

### Related JIRA tickets

SNT-553
SNT-546


Bewi (changes requested):
It works fine except for one edge case.
If you click on a connector, move it and attache it to the same node, the connection is lost anyway.

Bewi (approved):
LGTM

PR #352, opened 2026-07-28
SNT-550 Add CSV export button for metric values

Bewi:

Add an export data layer to CSV button

### Related JIRA tickets

SNT-550

## Changes

- Add an api endpoint to download a list of data layers as CSV for a given year (or none)
- Add a button on DataLayer header to download selected data layer as CSV

## How to test

Go to SNT Malaria, Data layers and try to download two type of data layers: 
Data layers with a given year (mostly custom layers) and data layers with no year (mostly from OH).
It should export the data layer as CSV and match the template format.

Once done, you can try to import that CSV for another year and see if it is the same.

## Print screen / video

image

## Notes

IASO PR https://github.com/BLSQ/iaso/pull/3228

## Doc

N/A

### DIFF
--- a/js/src/domains/dataLayers/dataLayerMap/ExportMetricValuesCsvButton.tsx
+++ b/js/src/domains/dataLayers/dataLayerMap/ExportMetricValuesCsvButton.tsx
@@ -24,7 +24,7 @@ export const ExportMetricValuesCsvButton: FC<Props> = ({
     const params = new URLSearchParams({
         metric_type_ids: metricType.id.toString(),
     });
-    if (year) {
+    if (year && year !== '0') {
         params.set('year', year);
     }
     // Use a plain anchor here, not bluesquare-components' IconButton/react-router Link:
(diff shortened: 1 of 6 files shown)


js/src/domains/dataLayers/dataLayerMap/ExportMetricValuesCsvButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (!metricType) {
+        return null;
+    }
+
+    const params = new URLSearchParams({
+        metric_type_ids: metricType.id.toString(),
+    });
+    if (year) {
+        params.set('year', year);
+    }
```
michiwend: when there's no dated data, year ends up as the string `'0'`, and since `'0'` is truthy in JS it gets sent to the API as `year=0`. It happens to work today because the backend's `Q(year=year) | Q(year__isnull=True)` filter always falls back to timeless values regardless... but it's a bit fragile/confusing. Might be cleaner to just omit the param when there's no real year selected?

michiwend (changes requested):
Overall LGTM. 
Regarding UX we should consider adding the download option also to the context menu on the list items (in addition). There is also one minor finding down in the comments.

Bewi:
> Overall LGTM. Regarding UX we should consider adding the download option also to the context menu on the list items (in addition). There is also one minor finding down in the comments.

I agree, however, problem with the context menu is for choosing a year for which you want to export your value.
I think it would be interesting to export for all years and to have a way to import for one single metric type and multiple years

michiwend (approved):

PR #353, opened 2026-08-04
SNT-556 Scenario reference year should be per data layer

Bewi:

Apply reference_year on data layer instead of full scenario

### Related JIRA tickets

SNT-556

## Changes

- Replace reference_year by a dictionary of data_layer_reference_year
- Still allow to select a reference year for the scenario but apply it to all layers
- Add inputs for reference year per data layer
- Refresh assignments when updating scenario reference years

## How to test

Try to create a scenario, under reference_year you should have a small toggle, which opens the per data layer section.
Try to set a global reference_year for which your data layer doesn't have value. 
Then set proper date for only one of the data layer.
On the scenario, create a rule, for the data layer with correct date, it should display it on the preview map if you select that layer and if you filter based on it, it should work as well.
For others, it should select nothing as we don't have any data for them.

Try to edit the reference_year, it should update assignments too.

## Print screen / video

image

## Notes

N/A

## Doc

N/A


js/src/domains/messages.ts
```diff
@@ -536,14 +536,49 @@ export const MESSAGES = defineMessages({
...
+            'Leave unset to use timeless values (or any year) by default.',
+    },
+    customizeDataLayerYears: {
+        id: 'iaso.snt_malaria.scenario.customizeDataLayerYears',
+        defaultMessage: 'Customize reference year per data layer',
+    },
+    dataLayerYearsLabel: {
+        id: 'iaso.snt_malaria.scenario.dataLayerYearsLabel',
+        defaultMessage: 'Data layer reference years',
+    },
```
michiwend: Never used

js/src/components/LayerSelect.tsx
michiwend: knit pick: Naming: since it no longer tracks its own selection, `initialSelection` isn't really 'initial' anymore

js/src/domains/scenarios/components/ScenarioModal.tsx
```diff
@@ -77,7 +82,7 @@ type ScenarioDialogProps = {
     onClose: (shouldRedirect: number | boolean) => void;
     scenario?: Scenario;
     titleMessage?: string;
-    operation: UseMutateFunction<unknown, unknown, ScenarioFormValues, unknown>;
+    operation: UseMutateFunction<unknown, unknown, unknown, unknown>;
```
michiwend, line 85: doesn't the payload still match `ScenarioFormValues` shape, why replacing it with `unknown`?

michiwend (approved):
Works well and LGTM except for those few knits. Feel free to address them

js/src/domains/scenarios/components/ScenarioModal.tsx
Bewi: There is a typing issue and we always get unknown, didn't find a solution for this

PR #354, opened 2026-08-04
Feature SNT-554: Consolidate chat, layers and node library

michiwend:

Consolidate chat, layers and node library. Last part of the UX improvements around the composite layers. 

### Related JIRA tickets

SNT-554
SNT-546

## Changes

- Consolidate chat, data layers and transformation nodes into the side panel using tabs to switch between chat and the new node library. The node library is a combined list of data layers and available transformation nodes
- Refactoring of the node builder and chat code
- Global fix for the sticky list headers that leaked the list items above
- Global re-styling of the scrollbars to remove the rail

## Print screen / video

localhost_8081_dashboard_snt_malaria_data-layers_accountId_2_(Small laptop)

localhost_8081_dashboard_snt_malaria_data-layers_accountId_2_(Small laptop) (1)


js/src/domains/compositeLayerEditor/hooks/useCanvasDrop.ts
```diff
@@ -83,19 +101,48 @@ export const useCanvasDrop = ({
...
-            id: newNodeId,
-            x: dropX,
-            y: dropY,
-            type: 'dataLayer',
-            width: 330,
-            connections: { inputs: {}, outputs: {} },
-            inputData: { metricType: { metricTypeId } },
-        };
-        shiftedNodes[newNodeId] = newNode;
+        if (nodeLibraryType === 'comment') {
```
Bewi, line 104: Is there a specific reason to not handle this like others ?
Size and colors ?

js/src/domains/compositeLayerEditor/nodeLibrary/NodeLibrary.tsx
```diff
@@ -0,0 +1,107 @@
...
+    metricCategories: MetricTypeCategory[];
+    /** Maps a MetricType id to the composite layer that produced it, when it is a composite. */
+    compositeLayerIdByMetricType: Map<number, number>;
+    /** The composite layer's own underlying layer, highlighted as in the browsing list. */
+    selectedMetricTypeId?: number;
+    /** Filter from `NodeLibrarySearch`, which lives in the card header. */
+    searchTerm: string;
+};
+
+const matches = (haystack: string, term: string): boolean =>
```
Bewi, line 20: Fact that I had to google translate haystack might mean that it is not the most obvious word :D 
Even though I salute originality

js/src/domains/compositeLayerEditor/nodeLibrary/nodeLibraryGroups.ts
```diff
@@ -0,0 +1,49 @@
...
+export type NodeLibraryGroup = {
+    label: string;
+    items: NodeLibraryItem[];
+};
+
+/**
+ * The library's own categories, above the data layer ones. Tools are canvas aids rather than graph
+ * nodes, so they have no `nodeTypeRegistry` entry.
+ */
+export const getNodeLibraryGroups = (
```
Bewi: For this kind of stuff I always wonder if it is not worth adding into a Memo, 'cause for now you use a memo on the caller, which is fine, but if we re-use that somewhere else, we might forget.

Bewi (approved):
Few comments, feel free to fix or ignore. 

Except from that, feature works like a charm.
LGTM

PR #356, opened 2026-08-05
bugfix SNT-560: fix inconsistent timeless data layers

michiwend:

Timeless data layers were saved with either `year=0` or `year=NULL`

### Related JIRA tickets

SNT-560

## Changes

Changes the all logic to use `year=NULL` for timeless metrics and migrates existing data with `year=0` to `year=NULL`


Bewi (approved):
LGTM

PR #358, opened 2026-08-05
SNT-523: consistent currency handling across the app

Bewi:

Explain here in one sentence.

### Related JIRA tickets

SNT-523

## Changes

Keep using defined currency as default, so if a user enter a value, we assume it is the value in the currency he configured.

- Centralize useGetBudgetSettings (we had multiple ones)
- Use currency from settings and display it where we have costs

## How to test

Go to settings, budget settings, change the currency (use a valid iso like EUR).
Check on intervention forms if it is displayed properly.
Check on budget totals if it is displayed properly.

## Print screen / video

image

image

image

## Notes

N/A

## Doc

N/A


michiwend (commented):
It works well but I struggle a bit with the UX

The fact that we have a free-text field and you can input anything that then gets mapped is a bit strange but I understand the decision from a the point of view that we can support whatever they enter. My bigger struggle is with how then the  configured currency is displayed in the inputs

Here USD becomes US$ and gets glued together with the amount. Its a bit hard to read and I would also expect to just have `$`, as it is already the symbol for USD, not `US$`. 
Image


The other thing is that for some currencies you write it before the amount and for others after the amount. Maybe instead of assuming this, we could make the currency just a hint or label inside the inputs? Something like this maybe?

Image

michiwend (approved):
LGTM and works really well

PR #362, opened 2026-08-10
Feature SNT-547: Year selection for data layer nodes

michiwend:

Allows to work either with yearly data or pin a data layer to one year

### Related JIRA tickets

SNT-547

## Changes

image

Adds a year selector to data layer nodes for layers with yearly data. A user can choose between two ways of working with yearly data now:
- The default selection is "Yearly" which is the previous behaviour. Yearly data is passed down the graph
- A single year is pinned. Passes only the data of that year down the graph. Equivalent to using a non-yearly data layer

A use case this unlocks is to compare yearly data of a data layer. A user can add the same data layer multiple times to the graph, each with a different year pinned. This way data of different years of the same data layer can be compared or combined.

This also adds the necessary changes to the API and system prompt so the chat has the context about how to work with years and which years are available for a given data layer. 

## Screenshots
Comparing two years of the same data layer

image


js/src/domains/compositeLayerEditor/hooks/useDataLayerYears.ts
```diff
@@ -0,0 +1,45 @@
...
+ * data, so this is computed once here and handed down through `CompositeEditorContext`.
+ *
+ * Shares its cache with `useGetMetricValues({ metricTypeId })` (same query key shape), so a node's
+ * own map preview doesn't re-fetch what this hook already loaded, and vice versa.
+ */
+export const useDataLayerYears = (
+    metricTypeIds: number[],
+): Map<number, number[]> => {
+    const results = useSnackQueries(
+        metricTypeIds.map(metricTypeId => ({
```
Bewi: Wouldn't it be better to accept a list of ids on metricvalues and return a dictionnary to prevent multi fetch ?

js/src/domains/compositeLayerEditor/utils/graph.ts
```diff
@@ -87,3 +88,49 @@ export const getConnectedDataLayerIds = (graph: FlumeGraph): number[] => {
...
+ * AI chat, or built by the canvas-drop handler bypass that and need it applied here instead.
+ */
+export const withDefaultSelectedYear = (
+    graph?: FlumeGraph,
+): FlumeGraph | undefined => {
+    if (!graph) return graph;
+    let changed = false;
+    const next: FlumeGraph = { ...graph };
+    Object.entries(graph).forEach(([nodeId, node]) => {
+        if (node?.type !== 'dataLayer') return;
```
Bewi: I think as we are using 'dataLayer' a lot, could be nice to have it defined in an enum or constant, with all others node types

Bewi (changes requested):
LGTM, only 1 optimisation when getting metric values

Bewi (approved):
LGTM

PR #364, opened 2026-08-11
Feature SNT-548: Add new org unit filter node and stack operation

michiwend:

Add new org unit filter node and stack operation to the combine node

### Related JIRA tickets

SNT-548

## Changes

Brings org unit filtering to the composite node builder. With the org unit filter node, any data layer can be filtered down to a set of hand-selected org units. The output of the filter node is the unchanged values of the input data layer but filtered to only the selected org units. The output data simply doesn't include any metrics for excluded org units.

Filtering is done in two modes "include selected" or "exlcude selected". Selection of org units is done either directly in the mini map inside the node or in the enlarged map dialogue which also provides a search bar for org units.

The second addition of this PR is a new "Stack" operation in our existing combine node. It allows to combine filtered data layers in a stacked manner which is similar to how our rule builder works with multiple rules. All stacked layers get merged from top to bottom with lower layers overriding districts of higher layers.

Additional changes are regarding AI chat integration of the new node and operation and some hardening of the LLM response parsing.

## Print screen / video
Example of using the filter node and stack operation to combine different districts from different layers
image

New stack operation (left) and enlarged district picker (right)
image image


js/src/domains/compositeLayerEditor/components/OrgUnitFilterControl.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    )
+                }
+            >
+                <option value="all">
+                    {formatMessage(MESSAGES.filterModeAll)}
+                </option>
+                <option value="none">
+                    {formatMessage(MESSAGES.filterModeNone)}
+                </option>
+            </select>
```
beygorghor, line 129: Is there a reason you didn't use Mui autocomplete here ?

js/src/domains/compositeLayerEditor/components/OrgUnitSelectionPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+ * portaled outside the canvas's `transform: scale(...)`, so plain MUI components are safe here
+ * (unlike the on-canvas control body).
+ */
+export const OrgUnitSelectionPanel: FC<Props> = ({
+    value,
+    onChange,
+    orgUnits,
+}) => {
+    const { formatMessage } = useSafeIntl();
+    const [search, setSearch] = useState('');
```
beygorghor: You can type the state value here => useState<string>

js/src/domains/compositeLayerEditor/components/OrgUnitSelectionPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+                                    edge="start"
+                                    checked={picked}
+                                    tabIndex={-1}
+                                    disableRipple
+                                />
+                            </ListItemIcon>
+                            <ListItemText primary={orgUnit.name} />
+                        </ListItemButton>
+                    );
+                })}
```
beygorghor: i'd prefer put this one in a separate component with his own scope

js/src/domains/compositeLayerEditor/components/StackPriorityControl.tsx
```diff
@@ -0,0 +1,107 @@
...
+        minWidth: 16,
+        textAlign: 'center',
+        color: 'text.secondary',
+        fontSize: 12,
+    },
+    portLabel: {
+        flex: 1,
+        fontSize: 13,
+    },
+} satisfies SxStyles;
```
beygorghor, line 22: const styles: SxStyles = .... should be enough here

js/src/domains/compositeLayerEditor/components/StackPriorityControl.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    </IconButton>
+                    <IconButton
+                        size="small"
+                        disabled={index === order.length - 1}
+                        onClick={() => move(index, 1)}
+                        aria-label={formatMessage(MESSAGES.stackMoveDown)}
+                    >
+                        <ArrowDownwardIcon fontSize="inherit" />
+                    </IconButton>
+                </Box>
```
beygorghor: Same here, if you loop into something i prefer a separate component

js/src/domains/compositeLayerEditor/compositeLayerChatBot/extractGraphSpec.ts
beygorghor, line 1: This file should unit tested !

js/src/domains/compositeLayerEditor/compositeLayerChatBot/extractGraphSpec.ts
```diff
@@ -68,13 +77,31 @@ export const extractGraphSpecFromFlume = (
...
+                          inputData.operation?.priorityOrder,
+                          entries.map(entry => entry.port),
+                      )
+                          .map(
+                              port =>
+                                  entries.find(entry => entry.port === port)
+                                      ?.nodeId,
+                          )
+                          .filter((id): id is string => id !== undefined)
+                    : entries.map(entry => entry.nodeId);
```
beygorghor, line 110: this is not really readable 😵

js/src/domains/compositeLayerEditor/utils/orgUnitSelection.ts
beygorghor, line 1: On this one two we should write unit tests

beygorghor (changes requested):
- If you can write some more frontend tests especially on files with small utils functions.
-  Default selection mode differs FE vs BE,  align default mode across UI, pydantic, and _resolve_selected_org_units.
UI DEFAULT_ORG_UNIT_SELECTION + normalizeSelection => "all"
Evaluator (config.get("mode") or "none") => "none"
Pydantic OrgUnitSelectionSpec => "none"
- LLM prompt size => Every chat turn gets the full district catalog via get_intervention_org_units. Fine for small accounts; risky for large ones (tokens, latency, truncation). Worth a short note in the PR (or a follow-up: cap, search, or only inject when the user mentions districts).
- Can you also fix eslint issues / warning please ?

js/src/domains/compositeLayerEditor/components/OrgUnitFilterControl.tsx
michiwend: mui popups portal to the document.body, which doesn't inherit the Flume canvas' CSS scale() transform, so they render at the wrong size/position when zoomed. I don't really like this solution either but it works...

michiwend:
@beygorghor created https://bluesquare.atlassian.net/browse/SNT-567 to fix the prompt size. Its gonna be a little tricky but I have some ideas.

